PR #46, opened 2017-05-03
Fixup SASS styles + misc

jimmcgaw:
Add npm start to scripts; remove bootstrap classes causing funny
busness; move throbber styles into component inline; update SASS layout,
remove flexbox from components.

### DIFF
--- a/src/components/general/Throbber.js
+++ b/src/components/general/Throbber.js
@@ -1,6 +1,17 @@
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
 
+const throbberDotStyles = {
+  'margin': '10px',
+  'height': '20px',
+  'width': '20px',
+  'borderRadius': '10px',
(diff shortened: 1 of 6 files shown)


src/components/general/Throbber.js
```diff
@@ -1,6 +1,17 @@
...
 
+const throbberDotStyles = {
+  'margin': '10px',
+  'height': '20px',
+  'width': '20px',
+  'borderRadius': '10px',
+  'backgroundColor': '#999',
+  'float': 'left',
+  'transform': 'scale(0.7)'
+}
```
elimisteve: Sweet

elimisteve:
@jimmcgaw No more gaps on the sides of the screen, yay! Thanks Jim.

Later we'll need to do ~100% flexbox so that it's compatible with React Native, but this is a lot better than what's there now. Just merged :+1:.

PR #54, opened 2017-05-06
Fixes #53 (partially)

### LINKED ISSUE #53: Break Apart Entry, Root Component, and Encryption
Right now a single file (src/index.js) contains the entry point for the React application, the default root layout component containing all other components, and a lot of logic that is doing encryption / decryption / processing of app data. 

Two problems here: 

1. Adding logic to the overall React application (that would normally go in the entry point) is tricky.
2. More critically, our JavaScript presentation library (React) is tied very heavily to application logic doing data processing. All data processing and calls to fetch external data sources should be moved into their own module and called by the React component as needed. This will make it easier to manage the layout component itself, or to move away from React in the future, if we want to.

jimmcgaw:
This change simplifies the entry point of the React application by
moving the root component code into its own file and importing.


jimmcgaw:
I'd like to merge this soon, so we can all be on the same page when editing the root component, instead of ending up with merge conflicts because much of the key code is being edited in src/index.js by someone, while someone else is editing src/components/App.js, provided, of course, that you approve of this change.

PR #56, opened 2017-05-07
Fix #55 Refactor root layout component

### LINKED ISSUE #55: Decouple Root Component from Business Logic
The root component of the application contains a lot of non-presentational logic, e.g. stuff that's doing encryption / decryption / processing of app data. This mean our JavaScript presentation library (React) is tied very heavily to application logic doing data processing.

All data processing and calls to fetch external data sources should be moved into their own modules and called by the React component as needed. This will make it easier to manage the layout component and the business logic, or to move away from React in the future, if we want to.

Some more specific goals:
* root component file does not import any encryption functions, like minilock stuff, which should live in its own module where possible.
* update functions to be arrow functions so we don't need closure `let that = this` stuff anywhere.
* ??

It's tricky because a lot of this code retrieves and sets the root component state. A move to redux or other flux architecture would simplify this, since state becomes more of a global object that can be read or modified from anywhere, instead of only in the context of the root component.

jimmcgaw:
This change cleans up the root layout component by using arrow functions
whenever the enclosing scope is needed by an anonymous function
definition, declaring functions as callbacks instead of anonymous ones,
re-ordering the function layout to be in the order in which chained
functions might be invoked, using ES6 template strings for
stringbuilding, and creates separate modules where possible.


jimmcgaw:
No new functionality, just organizing. Seems to work, but in the absence of test coverage I can't be sure something didn't break, say in an error handler callback, which typically don't get triggered. Also "receive message" flow is only testable if you've got other users sending you messages. Give it a try?

elimisteve:
> Also "receive message" flow is only testable if you've got other users sending you messages.

I've been talking to myself via LeapChat a lot lately :-)

jimmcgaw:
Haha, so very true.

src/utils/encrypter.js
```diff
@@ -5,30 +5,30 @@ const sha384 = require('js-sha512').sha384;
 const emailDomain = '@cryptag.org';
 
 export function getEmail(passphrase){
-	return sha384(passphrase) + emailDomain;
```
jimmcgaw: My editor keeps defying my orders to use spaces instead of tabs. :\

jimmcgaw:
LGTM.

elimisteve:
@jimmcgaw :+1: Merrrrrrged

PR #63, opened 2017-05-08
Fix #62: Wrap long continuous text lines

### LINKED ISSUE #62: Mobile bug: when non-whitespace-containing text overflows, shouldn't zoom out
> Guess we need text-overflow / text-wrap set properly.

jimmcgaw:


elimisteve:
:+1:

PR #70, opened 2017-05-10
When re-authing, set messages to []

### LINKED ISSUE #67: Bug - Page Reload duplicates messages
On leapchat.org, if the server goes down for any reason, the client properly displays the connection error message, but when it reconnects and reloads the messages, it will double-load several, if not all, of the previous messages.

screen shot 2017-05-09 at 7 26 18 am

elimisteve:
Fixes #67


elimisteve:
@jimmcgaw This solution isn't as good, methinks

PR #71, opened 2017-05-10
When first message on the messages WS received, set messages to []

### LINKED ISSUE #67: Bug - Page Reload duplicates messages

elimisteve:
Fixes #67

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -182,6 +182,7 @@ export default class App extends Component {
 
   newWebSocket(url){
     let ws = new WebSocket(url);
+    ws.firstMsg = true;
 
     ws.onopen = (event) => {
       let authToken = this.state.authToken;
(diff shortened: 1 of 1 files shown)


elimisteve:
@jimmcgaw This solution _looks_ jankier compared to https://github.com/cryptag/leapchat/pull/70 , but the duration of time where there are zero messages are on the screen is lower here, so this solution seems better to me.

Once we have persistence, let's create a new WS endpoint, `/api/ws/messages/since/<message_id>`, that lets the client ask for all messages since the last one they received.

jimmcgaw:
Looks good! Not sure how to repro locally, but this seems like it'll do the trick.

PR #79, opened 2017-05-12
Fix #74: See username in your own messages; username is bold instead …

### LINKED ISSUE #74: Your Own Messages Show Username
Right now it's assumed that you don't need to see your own username
displayed in messages that are styled as coming from you. Someone's
username might change if they randomly generate one, or they might
forget it, so show the username on all messages.

jimmcgaw:
…of italic; widen message box layout


elimisteve:
Excellent :+1:. Merged!

Issue #84, opened 2017-05-14
Display User list in sidebar


jimmcgaw:
On mobile, the current user list needs to be a list that expands when the users icon at the top left is clicked.

On desktop, the current user list needs to be a list that is diplayed in the sidebar, under the users icon.

elimisteve:
Want to create an empty component for me to populate with the user list?

jimmcgaw:
That was the intent of this: https://github.com/cryptag/leapchat/pull/85/files#diff-e7397b738e0b4d49855b9d97403d8c84

If that can't be merged before tackling this, I can create something else based on master.

elimisteve:
Finished with #31

PR #85, opened 2017-05-14
Fix #77: Responsive Header

### LINKED ISSUE #77: Create Header Menu
Create a header menu. Start with "mobile first" in mind and develop the mobile web version, keeping in mind that the desktop web version will be differently styled.

Key things:
- Top left: person icon which slide toggles a list of participants (depends on presence, #31)

- Middle: LeapChat logo (need asset for this)

- Top right: Gear icon that expands menu, with single item that allows user to bring up modal to set their username. Or the gear icon could just bring up username modal for now.

Consider desktop design. Sidebar with navigation links grouped by header?

Use styles (background color, text color) from the backchannel project for now.

jimmcgaw:
This change creates a responsive header for the app that lives at the
top of the site on mobile and as a sidebar for larger desktop screens.
It converts the primary elements in the layout to use the new flexbox
api. It also sets a fixed height on the messages container to allow for
scrolling when the message list overflows the height of the container,
which is less than flexy but will work for now. Elements are reordered
in the header between the mobile and desktop versions.

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -9,7 +9,6 @@ body {
 
 .encloser {
   display: flex;
-  min-height: 100vh;
   flex-direction: row; }
 
 @media (max-width: 768px) {
@@ -27,19 +26,20 @@ header {
   flex-direction: row;
   justify-content: space-between; }
   header .users-list {
-    order: -1; }
+    order: -1;
+    flex-grow: 1; }
     header .users-list .users-icon {
       cursor: pointer; }
     header .users-list ul {
       display: none; }
+  header .logo-container {
+    display: flex;
+    justify-content: space-between;
+    flex-grow: 1; }
   header .logo {
     text-align: center;
     font-size: 24px;
     font-weight: bold; }
-  header .menu-list {
-    display: none; }
-    header .menu-list i.fa {
-      margin-right: 5px; }
 
 @media (min-width: 768px) {
   header {
@@ -48,11 +48,16 @@ header {
     flex-direction: column;
     justify-content: start;
     height: 100vh; }
-    header .logo {
-      text-align: left;
-      order: -3; }
-    header .settings {
-      order: -2; } }
+    header .logo-container {
+      margin-bottom: 15px;
+      order: -1;
+      flex-grow: 0; }
+    header .users-list {
+      flex-grow: 0; }
+      header .users-list .users-icon {
+        display: none; }
+      header .users-list ul {
+        display: block; } }
 
 main {
   display: flex;
@@ -72,7 +77,6 @@ main {
         padding: 5px 10px;
         margin-bottom: 5px;
         width: 90%;
-        position: relative;
         word-wrap: break-word; }
         main .content .message-box ul li .username {
           display: block;
@@ -89,8 +93,6 @@ main {
         main .content .message-box ul li.chat-incoming {
           float: left;
           color: #333; }
-      main .content .message-box ul li:hover .delete-message {
-        visibility: visible; }
     main .content .message-list {
       padding: 0 15px; }
     main .content .message-form {
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -73,8 +73,7 @@ main {
         margin-bottom: 5px;
         width: 90%;
         position: relative;
-        word-wrap: break-word;
-        flex-grow: 1; }
+        word-wrap: break-word; }
         main .content .message-box ul li .username {
           display: block;
           font-weight: bold; }
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -7,68 +7,127 @@ body {
   font-size: 16px;
   font-family: 'Lato', sans-serif; }
 
-.container {
-  width: 100%; }
+.encloser {
+  display: flex;
+  min-height: 100vh;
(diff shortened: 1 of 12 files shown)


jimmcgaw:
This does not address the user list that needs to be displayed on mobile and desktop. For this: https://github.com/cryptag/leapchat/issues/84

jimmcgaw:
Updated. Also: https://github.com/cryptag/leapchat/issues/86

elimisteve:
@jimmcgaw Mobile looks perfect, nice! On larger screens, how about removing `fa-user` entirely, since the user list will always be visible, and how about putting `fa-cog` to the right of "LeapChat"? I think that'll look great and also free up more space below for the user list and, eventually, the channels list.

screenshot-leapchat-responsive-desktop-screenshot_2017-05-14_14-26-03

static/css/main.css
```diff
@@ -7,68 +7,127 @@ body {
   font-size: 16px;
   font-family: 'Lato', sans-serif; }
 
-.container {
-  width: 100%; }
+.encloser {
+  display: flex;
+  min-height: 100vh;
```
elimisteve: @jimmcgaw I'm not how to say this, but is `min-height: 100vh` "flexbox-friendly"?

jimmcgaw: Hrm, good catch. I'm not sure whether or not this is good form from a flexbox point of view, but it's actually unnecessary. Given the layout as it is, it will fill the viewport top to bottom without this. I'll remove.

static/css/main.css
```diff
@@ -7,68 +7,127 @@ body {
...
+    flex: 1;
+    min-width: 300px;
+    flex-direction: column;
+    justify-content: start;
+    height: 100vh; }
+    header .logo {
+      text-align: left;
+      order: -3; }
+    header .settings {
+      order: -2; } }
```
elimisteve: @jimmcgaw Interesting. I assume the default `order` is 0 or 1?

jimmcgaw: The default initial value is 0. Anything negative will be displayed before `order: 0;` in the main-axis, and anything positive after it.

static/css/main.css
```diff
@@ -7,68 +7,127 @@ body {
...
+      overflow-y: scroll;
+      padding-top: 5px; }
+      main .content .message-box ul li {
+        display: block;
+        border: solid 1px #eee;
+        border-radius: 10px;
+        padding: 5px 10px;
+        margin-bottom: 5px;
+        width: 90%;
+        position: relative;
```
elimisteve: @jimmcgaw What does this do?

static/css/main.css
```diff
@@ -7,68 +7,127 @@ body {
...
+  main .content {
+    display: flex;
+    flex-direction: column;
+    flex-grow: 1; }
+    main .content .message-box {
+      height: calc(100vh - 120px);
+      overflow-y: scroll;
+      padding-top: 5px; }
+      main .content .message-box ul li {
+        display: block;
```
elimisteve: @jimmcgaw What does this do?

static/css/main.css
jimmcgaw: If this is in reference to `position: relative`, I believe it was originally used to position the "delete message" button absolutely within the message element itself. That element was removed at some point, maybe in the transition from backchannel, so the answer here is: nothing. :)

static/css/main.css
jimmcgaw: By default `li` elements are `display: inline-list;`. I'm not sure of this, but I don't think you can float elements with that display property. In that case, this makes them floatable. If I'm wrong about that, then it's possible that this might not do anything.

jimmcgaw: Sorry, they're `display: list-item;` by default. Not `inline-list`.

static/sass/_layout.scss
```diff
@@ -5,99 +5,232 @@ html, body, ul {
...
+          /* this font-size disables zooming
+          on mobile safari, which breaks layout */
+          font-size: 16px;
+        }
+      }
+
+      button {
+        border: none;
+        padding: 5px;
+        float: right;
```
elimisteve: Doesn't make sense to just use flexbox here rather than floating the button, as to prevent the possibility of the button floating above other things, plus other reasons?

And if the button -- and especially the `textarea` above -- expanded to fill the available space using flexbox, then if the user zoomed out (`ctrl+-` or `ctrl+mousewheel down`), the textarea and the send button would still be touching, yes?

jimmcgaw: Yep, I totally agree. I chose not to address that here and now b/c https://github.com/cryptag/leapchat/commit/3adc82bea6fa8e945beff2adfd0f54cc6ba6106d

It's a good question. In this case the button doesn't have a border, so them touching wouldn't be a visual styling problem. If that happened, it would just increase the "hit zone" of the button all the way to the edge of the textarea, which would be, arguably, not a bad thing. In the commit above, however, the button has some padding in what is a fixed width box, while the textarea is in a variable width box. The two shouldn't touch.

jimmcgaw:
@elimisteve Did some cleanup of the styles from the feedback, and moved the settings icon up to top right corner on desktop.

PR #98, opened 2017-05-18
Fix #81: Versioned production build of assets

### LINKED ISSUE #81: Implement versioning for static assets
Once the app has been released into the wild we'll have browser caching
issues to contend with. We should version assets in production with the
latest and greatest mechanism, which might or might not be the hash
querystring param after the filename. Investigate best solution and add
it.

jimmcgaw:
This uses gulp and a few tools to create a single versioned, compressed file for each of our CSS and JS suitable for use in production.

Current bug: font-awesome is broken and not loading via link elements in index.html.

Tech debt: development workflow now includes these compressed files, without a source map, so debugging might be harder. We need to let development and production asset workflows live alongside each other. There's currently two index.html files, one for dev and one for production.

If font-awesome can be fixed, deploying this should result in a much faster load time and expire assets for users once they've been updated.


index.dev.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
```
jimmcgaw: Leaving this file here, since we'll want to get back to developing with this file that loads each thing independently.

index.html
```diff
@@ -6,35 +6,26 @@
...
+    <!-- inject:css -->
+    <link rel="stylesheet" href="/build/app-8a34b4f915.css">
+    <!-- endinject -->
+
+    <link href="static/lib/font-awesome/fonts/FontAwesome.otf">
+    <link href="static/lib/font-awesome/fonts/fontawesome-webfont.eot">
+    <link href="static/lib/font-awesome/fonts/fontawesome-webfont.svg">
+    <link href="static/lib/font-awesome/fonts/fontawesome-webfont.ttf">
+    <link href="static/lib/font-awesome/fonts/fontawesome-webfont.woff">
+    <link href="static/lib/font-awesome/fonts/fontawesome-webfont.woff2">
```
jimmcgaw: These are 404s. I'm not sure why, routing works for audio file below. Of course, this also assumes that 'bower install' is run in production environment and these files will be available at the specified path.

jimmcgaw: @elimisteve

elimisteve:
@jimmcgaw Instead of what I did here (namely include a modified `font-awesome.css` in git), I just learned that we should probably do https://github.com/Swiip/generator-gulp-angular/issues/337#issuecomment-73638794

elimisteve:
aka https://github.com/Swiip/generator-gulp-angular/issues/337#issuecomment-74403162

PR #99, opened 2017-05-18
Persistence

elimisteve:


README.md
```diff
@@ -58,14 +61,23 @@ To install and build static assets:
 
 ``` $ npm run dev  # watch asset files and recompile when changed ```
 
-Then, to build and run the go binary
+Then, to set up the database and run PostgREST, which our Go code uses for persistence:
+
+``` $ cd db/ ```
+
+``` $ sudo su postgres -c 'bash init_sql.sh' ```
```
jimmcgaw: FWIW, this didn't work on OS X. (Not saying this shouldn't be here, but it blocked me on getting started doing this.)

elimisteve: Hmm okay, good to know. The key is becoming the `postgres` user then running `bash init_sql.sh`.

pg_types.go
```diff
@@ -0,0 +1,129 @@
...
+		return fmt.Errorf(
+			"Got HTTP %v from PostgREST, wanted %v. Resp: %#v (err unmarshal: %v)",
+			resp.StatusCode, statusWanted, respjson, err)
+	}
+
+	return nil
+}
+
+func (cl *PGClient) Get(urlSuffix string) (*http.Response, error) {
+	log.Debugf("GET'ing from %s", urlSuffix)
```
jimmcgaw: Debug? Okay to leave this in?

jimmcgaw:
Looks good! I don't know go well enough to do a proper review, but I didn't see any massive red flags.

pg_types.go
elimisteve: A few weeks ago, @ajvb set up `logrus`, so we now have log levels. That statement will only print to stdout if we're not in production, aka when we're dev'ing (yaaay)

PR #117, opened 2017-05-27
Fixes #116: use a single index.html file which gets copied to build d…

### LINKED ISSUE #116: Combine index.html and index.dev.html
Optimizing dev workflow due to small nit

jimmcgaw:
…ir, then injected with production assets; server.go serves this file instead


elimisteve:
I don't fully understand why, but it looks fine to me :-) :+1:

PR #121, opened 2017-06-06
Added "overflow-wrap: break-word;", decreased width

### LINKED ISSUE #120: CSS bug: if message consists of long whitespaceless string, message box extends off the page + page zooms out on mobile

elimisteve:
Fixed #120, but now causes a float issue :-\


elimisteve:
@jimmcgaw Could you please check out this branch (in both senses of the term) and maybe somehow fix the CSS bug I introduced without getting rid of the one I fixed?

elimisteve:
@jimmcgaw The right-floated outgoing messages just need to be brought in further left, looks like

elimisteve:
@ajvb @jimmcgaw I'll be at Noisebridge today from 4pm till at least 10pm; I'm finally learning CSS and will try to fix the above, but any tips you have for why the width gets so messed up when long, whitespace-free messages are sent wolud be helpful, for example.

Lots of other important stuff, too: https://github.com/cryptag/leapchat/milestone/4 ; https://github.com/cryptag/leapchat/milestone/5

elimisteve:
@ajvb Let's look at this Saturday; once this is fixed I will immediately announce/launch LeapChat to a broader audience!

elimisteve:
Fixed by @MartyTheeMartian!!!

elimisteve:
I've cherry-picked these enhancements separately

PR #124, opened 2017-06-25
basic webpack

dradetsky:
probably don't want to straight up merge this, but it might be a useful starting point

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,51 @@
+const path = require('path')
+const webpack = require('webpack')
+const WebpackMd5Hash = require('webpack-md5-hash')
+const HtmlWebpackPlugin = require('html-webpack-plugin')
+const InlineManifestWebpackPlugin = require('inline-manifest-webpack-plugin')
+const ExtractTextPlugin = require('extract-text-webpack-plugin')
+
+const config = {
+  entry: {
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+        loader: 'babel-loader',
+        options: {
+          presets: ['react', 'es2015']
+        }
+      }
+    }]
+  }
+}
+
+module.exports = [config]
```
elimisteve: @dradetsky `config` should be in an array for some reason?

dradetsky: honestly, i can't remember why that's done

webpack.config.js
```diff
@@ -0,0 +1,51 @@
...
+const config = {
+  entry: {
+    vendor: './src/vendor.js',
+    main: './src/index.js',
+  },
+  output: {
+    path: path.join(__dirname, 'build'),
+    publicPath: '/build',
+    filename: '[name].[chunkhash].js',
+    chunkFilename: '[name].[chunkhash].js'
```
elimisteve: @Spetastian Seems like we need ^^this^^ for cache-busting?

elimisteve:
Thanks for this!  We used part of this for the current config :+1: which is working fantastically.

Thanks, gents!

PR #129, opened 2017-07-18
Use webpack

Spetastian:
Using webpack instead of gulp, bower and browserify. 

Instead of installing and bootstrapping static JS dependencies with bower and gulp, the various modules have either been moved to /src/static or /src/utils, or been replace with a npm module (e.g `bs58` for base58 encoding and decoding, `blake2s`, `bootstrap` etc.) and are imported as modules in /src/index.js or where it is used as a dependency since that is how webpack discovers modules that should be parsed, transpiled and stuffed into the bundle.

Also i have created modules out of `miniLock.js`, `nacl.js` and `scrypt.js`.


Spetastian:
@elimisteve I closed https://github.com/PursuanceProject/leapchat/pull/3 and opened the same PR here instead.

elimisteve:
@Spetastian Great work :+1: :tada:!

PR #132, opened 2017-07-20
fix Postgres install scripts for Mac OS

Moe-Shoman:


elimisteve:
@Moe-Shoman Merged! :+1: Thank you

PR #151, opened 2017-08-03
Define production env for production build

Spetastian:


elimisteve:
@Spetastian Works perfectly :+1:. Merged, and deployed!

PR #153, opened 2017-08-06
[WIP] Pin PostgreSQL install to 9.6

ajvb:


debian_install.sh
```diff
@@ -2,4 +2,7 @@
 # Steve Phillips / elimisteve
 # 2017.05.13
 
-sudo apt-get install postgresql-9.4 postgresql-contrib-9.4
+sudo add-apt-repository "deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main"
```
ajvb: :man_facepalming: this is `xenial` specific.

elimisteve: @ajvb Could use `$(grep DISTRIB_CODENAME /etc/lsb-release | cut -d= -f2)` or similar to dynamically get the Ubuntu codename, rather than hardcoding `xenial`?

A solution that works on Debian, too, would be ideal.

elimisteve: @ajvb Think that'd work?

elimisteve: Looks like it: http://apt.postgresql.org/pub/repos/apt/dists/

PR #160, opened 2017-08-17
CSS Fix for textarea

MartyTheeMartian:
textarea may change in height without displacing other divs


MartyTheeMartian:
@elimisteve PR fixes Last* CSS bug

PR #167, opened 2017-09-02
[DO NOT MERGE YET] Use redux observable

Spetastian:


src/epics/chatEpics.js
```diff
@@ -0,0 +1,227 @@
...
+    USER_STATUS_DELAY_MS
+} from '../constants/messaging';
+
+import { getEmail, getPassphrase, generateMessageKey } from '../utils/encrypter';
+import ChatHandler from './helpers/ChatHandler';
+import miniLock from '../utils/miniLock';
+import { combineEpics } from 'redux-observable';
+import createDetectVisibilityObservable from './detectPageVisibilityObservable';
+
+const BACKEND_URL = process.env.BACKEND_URL;
```
elimisteve: @Spetastian This is old and can be deleted; we're using `window.location.origin` instead so that LeapChat is accessible via multiple URLs at once (e.g., via a combination of an Onion Service (.onion site), localhost, proxies like ngrok, etc).

src/epics/chatEpics.js
```diff
@@ -0,0 +1,227 @@
...
+    action$.ofType(CHAT_SEND_MESSAGE)
+    .mergeMap(action =>
+
+        Observable.of(action)
+        .do(action => {
+            chatHandler.sendMessage(action.message, action.username);
+        })
+        .retryWhen(error => error.delay(1000))
+    )
+    .map(() => ({ type: 'XXXXX' }))
```
elimisteve: @Spetastian `CHAT_MESSAGE_SENT`, perhaps?

src/epics/detectPageVisibilityObservable.js
```diff
@@ -0,0 +1,36 @@
...
+      .map(() => {
+        if (document[hiddenKeyName]) {
+          return 'online';
+        } else {
+          return 'viewing';
+        }
+      }),
+    Observable.fromEvent(window, 'beforeunload')
+      .do(e => e.returnValue = 'Do you really want to quit?')
+      .map(() => 'offline')
```
elimisteve: @Spetastian So it makes sense to use `map` in these cases where there's only 1 value?  Usually I think of `map`ping over an array or the like... but should I be thinking about this in terms of a _stream_ of values?

elimisteve: each of which is being mapped to some other value with `.map(...)`?

src/epics/helpers/ChatHandler.js
```diff
@@ -0,0 +1,203 @@
...
+                    this.wsMessageSubject.next({
+                        fromUsername: message.tags.from,
+                        maybeSenderId: message.maybeSenderId,
+                        message: contents.msg
+                    })
+                })
+
+            reader.readAsText(message.fileBlob);
+            return chatMessageObservable;
+        } else {
```
elimisteve: @Spetastian Can you turn this into an `else if`, and then handle `else` by throwing an error saying "Unrecognized data with tags ..."?

src/static/lib/bootstrap/js/collapse.js
```diff
@@ -0,0 +1,212 @@
...
+    if (!$this.attr('data-target')) e.preventDefault()
+
+    var $target = getTargetFromTrigger($this)
+    var data    = $target.data('bs.collapse')
+    var option  = data ? 'toggle' : $this.data()
+
+    Plugin.call($target, option)
+  })
+
+}(jQuery);
```
elimisteve: @Spetastian Why include all this in the repo itself rather than `npm install`ing it?

src/epics/chatEpics.js
Spetastian: Yes, there are some of these placeholders left. I will fix them when i am doing the final touches.

src/epics/detectPageVisibilityObservable.js
Spetastian: Yes, you can se `.map(...)` in this way:
Map is an operator that transforms every value from a source into another value. For an array the source has a fixed set of values and the map call operates upon each and one of them.

In the case of an Observable stream, the map operator works the same but instead of a fixed set of values it can have an infinite amount of values pushed to the sources at different points in time.

In this specific case the consumer of this observable is not interested in the event object that the `Observabe.fromEvent(...)` emits. The only interesting values are the three states "online", "viewing" and "offline". 

Do you feel like this is a good explanation as to why we are using .map to map to a single value here?

src/epics/helpers/ChatHandler.js
Spetastian: That would be a good idea. What will happen when the error is thrown, how will it tell the user that something went wrong?

src/reducers/chatReducer.js
```diff
@@ -1,50 +1,103 @@
...
+
+    case CHAT_INIT_CONNECTION:
+      return Object.assign({}, state, {
+        messages: [],
+        connecting: true,
+        connected: false,
+        shouldConnect: false,
+        pincodeRequired: false,
+        ready: false,
+        usernameRequired: !state.username,
```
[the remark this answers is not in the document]
Spetastian: That is correct. This is just the state property that tells the application if the username modal should be visible or not: https://github.com/Spetastian/leapchat/blob/use-redux-observable/src/components/App.js#L128-L131

I know that the user should be able to open the modal manually and change the username via settings. I have not gotten to that specific point yet. That will probably change how this works.

src/epics/chatEpics.js
Spetastian: Ill fix it then 😄

src/static/lib/bootstrap/js/collapse.js
Spetastian: I actually dont now and i will look over it.

src/epics/chatEpics.js
elimisteve: :+1:

src/epics/helpers/ChatHandler.js
elimisteve: Using `displayAlert` sounds good to me

src/epics/chatEpics.js
elimisteve: :+1:

src/reducers/chatReducer.js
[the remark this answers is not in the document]
Spetastian: @elimisteve Does that sound reasonable to you or should we take another approach?

PR #168, opened 2017-09-03
Change port typo

iamrobinpham:
This is a low hanging fruit change!

Just set this up on my Arch device today and here are some notes that would be useful for a bigger refactor of the `README.md` file.
1. In the Docker setup section, setting up the 'db scripts' was a bit unclear, and didn't realize I had to use the `/db/sql/*.sql` files
2. For newer people, it's not so clear that you still have to run `npm install`, `npm run dev`, `go get`, `go build`, and `npm run be` when walking through the Docker setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,10 @@ containers:
 
 - Postgres at port 5432
 - PostgREST at port 3000
-- Adminer at port 8081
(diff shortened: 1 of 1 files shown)


elimisteve:
Merged. Thanks!

Issue #65, opened 2017-05-08
Messages show timestamp


jimmcgaw:
Messages should show the time they were sent if date matches
current, else just date for past days.

jimmcgaw:
Tricky part here is datetime stamps are in GMT, need to detect browser timezone and adjust time, possibly with something like https://www.npmjs.com/package/jstz

elimisteve:
@jimmcgaw A good time to create/record/pass-to-clients the timestamps is when implementing persistence; Postgres can give each message a timestamp, and also a UUID (which we can use as React keys rather than the strange timestamp-plus-index-string thing I concocted that we're currently using).

elimisteve:
PR for this that I wasn't sure about (efficiency? will React re-render every message every second as the relative timestamps change?): https://github.com/cryptag/leapchat/pull/169

PR #169, opened 2017-09-03
Simple relative timestamp solution

robin-pham:
#65

Adds timestamp to message. It's a barebones solution that can be refactored later if tooltips are used.
It uses [this library](https://date-fns.org/v2.0.0-alpha.2/docs/formatRelative).

Examples:
screenshot_2017-09-02_20-47-21
screenshot_2017-09-02_20-46-51
screenshot_2017-09-02_20-45-49

First time working in this codebase.... it's pretty funky, message keys are the timestamps whaaaat!

### DIFF
--- a/src/components/chat/Message.js
+++ b/src/components/chat/Message.js
@@ -1,4 +1,6 @@
 import React, { Component } from 'react';
+import { format, formatRelative } from 'date-fns';
+
 import emoji from '../../utils/emoji_convertor';
 import md from '../../utils/link_attr_blank';
 
@@ -7,6 +9,9 @@ class Message extends Component {
     let { message, username } = this.props;
     let fromMe = message.from === username;
     let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
+    
+    let RFCDate = message.key.split('-')[0];
+    let date = formatRelative(Date.parse(RFCDate), new Date());
 
     let emojified = emoji.replace_colons(message.msg);
 
@@ -29,9 +34,13 @@ class Message extends Component {
     // Render escaped HTML/Markdown
     let linked = md.render(emojiMD);
 
+
     return (
       <li className={'chat-message ' + messageClass} key={message.key}>
-        <span className="username">{message.from}</span>
+        <div className="message-header">
+          <span className="username">{message.from}</span>
+          <span className="message-timestamp">{date}</span>
+        </div>
         <div dangerouslySetInnerHTML={{__html: linked}}>
         </div>
       </li>
--- a/src/components/chat/Message.js
+++ b/src/components/chat/Message.js
@@ -11,7 +11,7 @@ class Message extends Component {
     let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
     
     let RFCDate = message.key.split('-')[0];
-    let date = formatRelative(Date.parse('Fri, 01 Sep 2016 03:42:50 GMT'), new Date());
+    let date = formatRelative(Date.parse(RFCDate), new Date());
 
     let emojified = emoji.replace_colons(message.msg);
 
--- a/src/components/chat/Message.js
+++ b/src/components/chat/Message.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { format, formatRelative } from 'date-fns';
+import { formatRelative } from 'date-fns';
 
 import emoji from '../../utils/emoji_convertor';
 import md from '../../utils/link_attr_blank';
@@ -11,7 +11,7 @@ class Message extends Component {
     let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
     
     let RFCDate = message.key.split('-')[0];
-    let date = formatRelative(Date.parse(RFCDate), new Date());
+    let date = formatRelative(Date.parse('Fri, 01 Sep 2016 03:42:50 GMT'), new Date());
 
     let emojified = emoji.replace_colons(message.msg);
 
@@ -33,7 +33,6 @@ class Message extends Component {
 
     // Render escaped HTML/Markdown
     let linked = md.render(emojiMD);
-
 
     return (
       <li className={'chat-message ' + messageClass} key={message.key}>
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "babel-loader": "^7.1.1",
     "btoa": "^1.1.2",
     "copy-webpack-plugin": "^4.0.1",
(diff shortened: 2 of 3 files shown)


elimisteve:
Hi @iamrobinpham, thanks for this! Checking it out...

src/components/chat/Message.js
```diff
@@ -7,6 +9,9 @@ class Message extends Component {
     let { message, username } = this.props;
     let fromMe = message.from === username;
     let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
+    
+    let RFCDate = message.key.split('-')[0];
```
elimisteve: @iamrobinpham FYI: soon, a timestamp will come from the server and we won't need to make assumptions about the `key`.

iamrobinpham: Sweet, I'll modify the PR once that happens

src/components/chat/Message.js
```diff
@@ -7,6 +9,9 @@ class Message extends Component {
...
+    let date = formatRelative(Date.parse('Fri, 01 Sep 2016 03:42:50 GMT'), new Date());
```
elimisteve: @iamrobinpham Was this supposed to be `formatRelative(Date.parse(RFCDate), new Date())`, or perhaps something else using `RFCDate`?

iamrobinpham: oops yeah! you got it `formatRelative(Date.parse(RFCDate), new Date())` is correct. I must've committed test code :(.

elimisteve: Okay, feel free to edit, commit the fix, and do a forced push to update this branch :-)

src/components/chat/Message.js
```diff
@@ -7,6 +9,9 @@ class Message extends Component {
     let { message, username } = this.props;
     let fromMe = message.from === username;
     let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
+    
+    let RFCDate = message.key.split('-')[0];
+    let date = formatRelative(Date.parse(RFCDate), new Date());
```
elimisteve: @Spetastian This will create a new `Date` object for every message on every render rather than creating a `Date` object once for each message _ever_ -- rather than on every render -- and attaching it to that message.

So we can create a new `Date` upstream and attach it to each message, or...

I'm thinking we just wait till after Nov 4th to integrate this in, since shortly thereafter we'll be getting a timestamp from the server anyway.  But using the UI that @iamrobinpham made should work well :-).

Spetastian: @elimisteve Did you meant to ping me here?

elimisteve: Yes, since I thought that you'd be the right person to attach the `Date` to each message... but just waiting a bit longer on this is probably better.

Thoughts?

Spetastian: Yeah, i would suggest that the date field is created in ChatHandler when parsing the message. So that this component just uses that property as is, no questions asked.

Spetastian:
@iamrobinpham The message component can be a function component instead of a class component since all it has is a render function. 

Instead of this:
```javascript
class Message extends Component {
  render(){
    let { message, username } = this.props;
    let fromMe = message.from === username;
    let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
    
    let RFCDate = message.key.split('-')[0];
    let date = formatRelative(Date.parse(RFCDate), new Date());

    let emojified = emoji.replace_colons(message.msg);

    // Convert `emoji.replace_colons`-generated <span> tags to Markdown
    let emojiMD = emojified.replace(
      /<span class="emoji emoji-sizer" style="background-image:url\((\/static\/img\/emoji\/apple\/64\/)(.*?)(\.png)\)" data-codepoints="(?:.*?)"><\/span>/g,
      (match, $1, $2, $3) => {
        // Example:
        //
        // $1 == /static/img/emoji/apple/64/
        // $2 == 1f604
        // $3 == .png
        // emoji.data[$2][3][0] == smile
        // return '![:smile:](/static/img/emoji/apple/64/1f604.png)'
        const twoSimple = ($2).split('-')[0];
        return '![:' + emoji.data[twoSimple][3][0] + ':](' + $1 + $2 + $3 + ')';
      }
    )

    // Render escaped HTML/Markdown
    let linked = md.render(emojiMD);

    return (
      <li className={'chat-message ' + messageClass} key={message.key}>
        <div className="message-header">
          <span className="username">{message.from}</span>
          <span className="message-timestamp">{date}</span>
        </div>
        <div dangerouslySetInnerHTML={{__html: linked}}>
        </div>
      </li>
    );
  }
}
```

You can do this:
```javascript
@iamrobinpham The message component can be a function component instead of a class component since all it has is a render function. 

Instead of this:
```javascript
  const message = ({ message, username }) => {
    let fromMe = message.from === username;
    let messageClass = fromMe ? 'chat-outgoing' : 'chat-incoming';
    
    let RFCDate = message.key.split('-')[0];
    let date = formatRelative(Date.parse(RFCDate), new Date());

    let emojified = emoji.replace_colons(message.msg);

    // Convert `emoji.replace_colons`-generated <span> tags to Markdown
    let emojiMD = emojified.replace(
      /<span class="emoji emoji-sizer" style="background-image:url\((\/static\/img\/emoji\/apple\/64\/)(.*?)(\.png)\)" data-codepoints="(?:.*?)"><\/span>/g,
      (match, $1, $2, $3) => {
        // Example:
        //
        // $1 == /static/img/emoji/apple/64/
        // $2 == 1f604
        // $3 == .png
        // emoji.data[$2][3][0] == smile
        // return '![:smile:](/static/img/emoji/apple/64/1f604.png)'
        const twoSimple = ($2).split('-')[0];
        return '![:' + emoji.data[twoSimple][3][0] + ':](' + $1 + $2 + $3 + ')';
      }
    )

    // Render escaped HTML/Markdown
    let linked = md.render(emojiMD);

    return (
      <li className={'chat-message ' + messageClass} key={message.key}>
        <div className="message-header">
          <span className="username">{message.from}</span>
          <span className="message-timestamp">{date}</span>
        </div>
        <div dangerouslySetInnerHTML={{__html: linked}}>
        </div>
      </li>
    );
  }

```
```

src/components/chat/Message.js
Spetastian: Actually, you should use `const` instead of `let` if you know that the variable will not be reassigned or redeclared which is true for basically all of the declarations in the render function.

src/components/chat/Message.js
Spetastian: @elimisteve Does that make sense?

PR #173, opened 2017-09-18
Implemented support for redux-observable with RxJS

Spetastian:


src/components/App.js
```diff
@@ -665,17 +159,17 @@ class App extends Component {
...
 const mapDispatchToProps = (dispatch) => {
   return {
-    addNewMessage: ({ key, fromUsername, maybeSenderId, message }) =>
-      dispatch(addMessage({ key, fromUsername, maybeSenderId, message })),
-    addNewUserStatus: ({ fromUsername, userStatus, created }) =>
-      dispatch(addUserStatus({ fromUsername, userStatus, created })),
-    clearAllMessages: () => dispatch(clearMessages()),
-    setUsername: (username) => dispatch(setUsername(username))
+    sendMessage: ({ message, username }) =>
+    dispatch(sendMessage({ message, username })),
```
elimisteve: Indent?

src/components/App.js
```diff
@@ -665,17 +159,17 @@ class App extends Component {
...
+    initConnection: (pincode) => dispatch(initConnection(pincode)),
+    initApplication: () => dispatch(initApplication()),
+    setUsername: (username) => dispatch(setUsername(username)),
+    dismissAlert: () => dispatch(dismissAlert())
   };
 };
```
elimisteve: @Spetastian Awesome work man! `App.js` looks fantastically clean.

elimisteve: @MartyTheeMartian You mentioned to me in another context that eliminating all state from the `App` component is desirable. Do you think so in this case?

src/components/chat/UserList.js
```diff
@@ -11,80 +11,69 @@ class UserList extends Component {
...
-    console.log('Updating presence/all user statuses:', viewing, online, offline);
-
     return (
       <div className="users-list">
         <ul style={this.styleUserList()}>
-          {viewing.map(status => {
+          {viewing.map((username, i) => {
             return (
-              <li key={status.key}>
+              <li key={i}>
```
elimisteve: @Spetastian Do keys not have to be globally unique? just unique to the component they're used within?

elimisteve: From the docs:

> The best way to pick a key is to use a string that uniquely identifies a list item among its siblings. Most often you would use IDs from your data as keys: ...

> When you don't have stable IDs for rendered items, you may use the item index as a key as a last resort: ...

Awesome, didn't know that :-).

src/constants/messaging.js
```diff
@@ -3,4 +3,4 @@ export const AUTH_ERROR = "Error authorizing you"; // Must match Go's miniware.A
 export const ON_CLOSE_RECONNECT_MESSAGE = "Message WebSocket closed. Reconnecting...";
 export const ONE_MINUTE = 60 * 1000;
 
-export const USER_STATUS_DELAY_MS = 10 * ONE_MINUTE;
+export const USER_STATUS_DELAY_MS = 2 * ONE_MINUTE;
```
elimisteve: @Spetastian We're going to have thousands of people using this. And right now, the LeapChat Go code sends the most recent 50 "messages" to the client, where a message could be a user status _or_ a useful message. So even though a user's status will be out of date for an extra 8 minutes by having `USER_STATUS_DELAY_MS` set to 10 minutes rather than 2 minutes, that'll reduce load on our servers considerably, and also show more useful messages right when users connect.

Soon enough we'll have real accounts people can optionally sign into, in which case the server will know which WebSocket connection corresponds to which user, and thus have the server send a message to the relevant chat room users _right_ when a user disconnects.  That requires having the clients hand over a bit more responsibility/trust to the server, but not much, and it'll make this whole user status thing much more efficient.

src/epics/chatEpics.js
```diff
@@ -0,0 +1,210 @@
...
+            })
+            .retryWhen(error => error.delay(500))
+    )
+    .map(username => usernameSet(username))
+
+const ownUserStatusEpic = (action$, store) =>
+    action$.ofType(CHAT_INIT_CHAT)
+    .mergeMap(() =>
+        createDetectVisibilityObservable()
+        //.throttleTime(USER_STATUS_DELAY_MS)
```
elimisteve: @Spetastian Should we throttle here?  I'm not sure what triggers `ownUserStatusEpic` to run, so I'm not sure.

src/epics/helpers/ChatHandler.js
```diff
@@ -0,0 +1,206 @@
...
+        this.wsMessageSubject = null;
+        this.wsUserStatusSubject = null;
+    }
+
+    onWsClose = (event) => {
+        this.wsMessageSubject.error(new Error('The websocket connection was closed!'));
+    }
+
+    onWsError = (event) => {
+        this.wsMessageSubject.error(new Error('Error occured on websocket connection!'));
```
elimisteve: @Spetastian Want to at least log the actual error?

src/reducers/chatReducer.js
```diff
@@ -1,50 +1,93 @@
...
 
+const PARANOID_USERNAME = ' ';
+const USERNAME_KEY = 'username';
+
 const initialState = {
+  connecting: false,
+  connected: false,
+  shouldConnect: false,
+  pincodeRequired: document && document.location.hash.endsWith('--'),
+  paranoidMode: document && document.location.hash.endsWith('---'),
```
elimisteve: Four dashes: `----`

src/reducers/chatReducer.js
```diff
@@ -1,50 +1,93 @@
...
+
+    case CHAT_INIT_CONNECTION:
+      return Object.assign({}, state, {
+        messages: [],
+        connecting: true,
+        connected: false,
+        shouldConnect: false,
+        pincodeRequired: false,
+        ready: false,
+        usernameRequired: !state.username,
```
elimisteve: @Spetastian So `usernameRequired` is used to determine whether we should ask the user for a username?  Could call it `askForUsername`?

src/components/App.js
```diff
@@ -44,618 +40,116 @@ class App extends Component {
...
-   * Closes the alert.
-   */
-  onAlertDismiss() {
-    this.setState({
-      alertMessage: ''
-    });
-  }
-
-  promptForUsername() {
+  handleShowSettings = () => {
```
elimisteve: @Spetastian :+1: for thinking ahead; we'll definitely add more settings soon.

src/epics/chatEpics.js
```diff
@@ -0,0 +1,210 @@
...
+import 'rxjs/add/operator/retryWhen';
+import 'rxjs/add/observable/of';
+import 'rxjs/add/observable/from';
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/observable/merge';
+import 'rxjs/add/observable/if';
+
+function createKeyPairObservable({ pincode = '' }) {
+    /* Using a function */
+    const keyPairObservable = Observable.create(function(observer) {
```
Moe-Shoman: what is the difference between declaring an observable then returning it, versus just directly returning like you did on line 87

src/components/chat/UserList.js
Spetastian: Best thing would be if we hade consistent unique IDs for messages to use. To generate a unique identifier as we did before on the client every time message is received is basically the same as using the index as a key.

src/constants/messaging.js
Spetastian: Of course, ill revert it.

src/epics/chatEpics.js
Spetastian: There is not much more difference than returning any other anonymous object vs. creating and assigning it first.

This is just me being inconsistent 😅

src/reducers/chatReducer.js
Spetastian: Well its not a function, its state. IMO It just represent the fact that a username is required since there are none. Then what the application does with that piece of state is something else that the Reducer does not care about. 

If we were to name it e.g `askForUsername` then that name represents expectations of how that piece of state might or might not be used in a completely different part (concern) of the application stack.

Just sharing how i think about it. I can rename it if you disagree.

src/components/chat/UserList.js
elimisteve: Each message has a UUID that we will soon expose to the frontend :-)

src/reducers/chatReducer.js
elimisteve: I see.  The ambiguity that I'm trying to figure out how to remove is: if "a username is required" is true, does that mean that we need a username, or that we both need a username _and we don't have one (yet)_?

src/epics/helpers/ChatHandler.js
```diff
@@ -0,0 +1,208 @@
...
+
+      for (let i = 0; i < binStrLength; i++) {
+        array[i] = binStr.charCodeAt(i);
+      }
+      const msg = new Blob([array], { type: 'application/octet-stream' });
+      miniLock.crypto.decryptFile(msg,
+        mID,
+        secretKey,
+        (fileBlob, saveName, senderID) => {
+          // NOTE: One tag can have several values. Talk to Steven
```
elimisteve: @Spetastian Let's tweak this post-merge

src/index.js
```diff
@@ -2,17 +2,20 @@ import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
 import App from './components/App';
 import { Provider } from 'react-redux';
-import { createStore, compose } from 'redux';
+import { createStore, compose, applyMiddleware } from 'redux';
+import { createEpicMiddleware } from 'redux-observable'
 import rootReducer from './reducers';
+import rootEpic from './epics';
 import ReduxDevTools from './components/_dev/ReduxDevTools';
 import 'isomorphic-fetch';
```
elimisteve: @Spetastian What does this do?

Spetastian: It imports whatever epics/index.js exports which is all the epics combined into one root epic. We only have one epic atm but this will be the file were we add new ones after we have created them in the epics folder.

Spetastian: It imports whatever epics/index.js exports which is all the epics combined into one root epic. We only have one epic atm but this will be the file were we add new ones after we have created them in the epics folder.

Spetastian: This is not suppose to be there anymore. But basically `fetch` is just a browser feature not a JavaScript feature. So if you want to write tests for that or run it on the server you will have to use one of these `isomorphic` modules that uses fetch in the browser but polyfills it if it does not exist.

Since we are using RxJS Ajax feature we do not need this at all.

src/epics/helpers/ChatHandler.js
Spetastian: That will be my next thing to look into then 😄

src/reducers/chatReducer.js
Spetastian: This is the initial state and originally it initially checks if the username is set and if is not, it shows the the modal where the user inputs the username.

Maybe we should just skip this propert and use the username directly. Instead of `const displaySettings = showUsernameModal || usernameRequired;` we just do `const displaySettings = showUsernameModal || username;` and then use displaySettings to decide if the modal is to be shown? Because if the username is ever not set we want the user to set it. This will naturally just happen once in the beginning of the application and after that we have a username set.

src/components/App.js
Spetastian: I dont think its a bad thing to have both Redux state and some internal component state. If all it does is to modify that specific components rendering but is not dependent on any external data or asynchronous operation or affects any sibling component then we could just use `this.setState(..)` and `this.state`.

But its more of a consideration from case to case. We dont want our components to be bloated.

@MartyTheeMartian What are your opinions about that?

src/epics/helpers/ChatHandler.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  onWsClose = (event) => {
+    console.error('Websocket got close event', event);
+    this.wsMessageSubject.error(new Error('The websocket connection was closed!'));
+  }
+
+  onWsError = (event) => {
+    console.error('Websocket got error event', event);
+    this.wsMessageSubject.error(new Error('Error occured on websocket connection!'));
```
elimisteve: @Spetastian `occurred` :-)

src/epics/helpers/ChatHandler.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  onWsMessage = (event) => {
+    const data = JSON.parse(event.data);
+    if (data && data.ephemeral && data.ephemeral.length && data.ephemeral.length > 0) {
+      Observable.from(data.ephemeral)
+        .mergeMap(ephemeral =>
+          this.createDecryptEphemeralObservable({ ephemeral, mID: this.mID, secretKey: this.secretKey }))
+        .mergeMap(this.resolveIncomingMessageTypeObservable)
+        .catch(error => console.error('An error occured in ChatHandler', error))
```
elimisteve: @Spetastian `occurred` :-) (with 2 `r`s)

Spetastian: Pfft, try to learn swedish 😉

Spetastian: Just kidding, will fix that 😄

src/reducers/chatReducer.js
Spetastian: @elimisteve What do you think about that?

src/epics/helpers/ChatHandler.js
elimisteve: Perfect

src/reducers/chatReducer.js
elimisteve: @Spetastian That sounds fantastic; love the simplicity -- great idea!

elimisteve:
@Spetastian Just tried this out and it works great!  Make the final `requireUsername` tweak, and I'll merge :-D

src/reducers/chatReducer.js
elimisteve: Make this tweak and I'll merge :+1:

src/reducers/chatReducer.js
```diff
@@ -1,50 +1,96 @@
...
+    case CHAT_SET_USER_STATUS:
+      return Object.assign({}, state, {
+        statuses: Object.assign({}, state.statuses, { [action.username]: action.status })
+      });
+
+    case CHAT_USERNAME_SET:
+      return Object.assign({}, state, {
+        username: state.paranoidMode ? PARANOID_USERNAME : action.username,
+        status: 'viewing',
+        usernameRequired: false,
```
elimisteve: @Spetastian Old, dead code?

elimisteve:
@Spetastian I just tried to build and got this error:

```
ERROR in ./src/epics/chatEpics.js
Module not found: Error: Can't resolve './helpers/createDetectPageVisibilityObservable' in '/home/user/gocode/src/github.com/cryptag/leapchat/src/epics'
 @ ./src/epics/chatEpics.js 25:44-101
 @ ./src/epics/index.js
 @ ./src/index.js
```

Spetastian:
@elimisteve Sorry for the inconvenience. I forgot to do git add after i moved a file. Also i removed the unnecessary property in the reducer.

src/reducers/chatReducer.js
```diff
@@ -1,50 +1,86 @@
...
+
 const initialState = {
-  username: '',
+  connecting: false,
+  connected: false,
+  shouldConnect: true,
+  pincodeRequired,
+  paranoidMode,
+  username: paranoidMode ? PARANOID_USERNAME : '',
+  previousUsername: localStorage && localStorage.getItem(USERNAME_KEY) || '',
```
elimisteve: @Spetastian Sweet :-)

src/epics/chatEpics.js
```diff
@@ -0,0 +1,212 @@
...
+              chatHandler.getUserStatusSubject()
+                .map(setUserStatus),
+
+              Observable.from([connectionInitiated(), alertSuccess('Connected.')])
+            ))
+
+      ).catch(error => {
+        console.error(error);
+        return Observable.from([
+          alertWarning('Lost connectiong to chat server. Trying to reconnect...'),
```
elimisteve: @Spetastian `connectiong` -> `connection` without the `g`

src/epics/chatEpics.js
```diff
@@ -0,0 +1,212 @@
...
+          )
+          .mergeMap(() =>
+            Observable.merge(
+              chatHandler.getMessageSubject()
+                .map(addMessage),
+
+              chatHandler.getUserStatusSubject()
+                .map(setUserStatus),
+
+              Observable.from([connectionInitiated(), alertSuccess('Connected.')])
```
elimisteve: @Spetastian Unfortunately `alertSuccess('Connected.')` shows up like ~50ms after `Observable.of(alertSuccess('New room created.'))`, which results in `New room created` not being seen.

How about removing the `Connected.` message until we change how Alerts work so that multiple can be showing at once?

src/epics/chatEpics.js
elimisteve: Reconnecting works great! Well done :+1:

src/components/App.js
```diff
@@ -1,37 +1,28 @@
 import React, { Component } from 'react';
-import $ from 'jquery';
 import miniLock from '../utils/miniLock';
```
elimisteve: @Spetastian You can remove this `import` line, too!

src/components/App.js
```diff
@@ -44,618 +35,119 @@ class App extends Component {
...
-    }
-
-    if (!this.paranoidMode()) {
-      localStorage.setItem(USERNAME_KEY, username);
-    }
-
-    this.props.setUsername(username);
-
+  handleSetUsername = (username) => {
     this.setState({
```
elimisteve: @Spetastian `previousUsername`'s value is indeed now being grabbed from localStorage, but the user's username isn't being stored there in the first place :-).

Would this line be a good place to do `localStorage.setItem(USERNAME_KEY, username)`, or I suppose that should be in `setUsername`...

src/epics/chatEpics.js
```diff
@@ -0,0 +1,212 @@
...
+    )
+    .catch(error => {
+      console.error(error);
+      return Observable.from([
+        alertWarning('Something went wrong when initiating. Trying again...'),
+        disconnected()
+      ])
+    });
+
+const setUsernameEpic = (action$, store) =>
```
elimisteve: @Spetastian Or maybe around _here_ is where `localStorage.setItem(USERNAME_KEY, username)` should be called?

src/components/App.js
elimisteve: @Spetastian According to my understanding, @MartyTheeMartian wasn't saying that mixing some local state and redux state is bad in general, but that having any local state in `App` in particular means that any time any of that state changes, the entire app will re-render, because `App` _and all its subcomponents will re-render_, aka the entire application.

src/epics/chatEpics.js
```diff
@@ -0,0 +1,212 @@
...
+  USERNAME_KEY
+} from '../constants/messaging';
+
+import { getEmail, getPassphrase, generateMessageKey } from '../utils/encrypter';
+import ChatHandler from './helpers/ChatHandler';
+import miniLock from '../utils/miniLock';
+import { combineEpics } from 'redux-observable';
+import createDetectVisibilityObservable from './helpers/createDetectPageVisibilityObservable';
+
+const BACKEND_URL = window.location.origin;
```
elimisteve: @Spetastian Can delete this line, too

src/epics/chatEpics.js
Spetastian: Good idea. Also this Rx flow might need som refactoring as well, but it works really fine for now.

src/epics/chatEpics.js
```diff
@@ -0,0 +1,212 @@
...
+  action$.ofType(CHAT_SET_USERNAME)
+    .filter(action => action.username && !store.getState().chat.paranoidMode)
+    .mergeMap(action =>
+      Observable.of(action.username)
+        .do(username => {
+          const ttl = USER_STATUS_DELAY_MS / 1000;
+          chatHandler.sendUserStatus(username, 'viewing', ttl)
+        })
+        .retryWhen(error => error.delay(500))
+        .do(() => localStorage && localStorage.setItem(USERNAME_KEY, action.username))
```
Spetastian: @elimisteve Correct, here it is. Here it saves to localStorage.

src/components/App.js
Spetastian: Yes, that is a thing to consider. That is why i dont not want App.js to be the component that connects to the Redux State and propagates it to all its children. Instead it just composes some of the "page" or "area" components that each connects to Redux and resolves the part of the state used in that context.

Also we should use Reacts `PureComponent` more in order to prevent unnecessary re-renders.

src/components/App.js
Spetastian: Since its a side effect that talks to the host OS, that is a part of the Rx flow in `chatEpics`

elimisteve:
@Spetastian @Moe-Shoman @MartyTheeMartian Sebastian's redux mega-branch: MERRRRRGED!!!!!! :tada: :tada: :tada: :-D :tada: :tada: :tada:

PR #181, opened 2017-10-05
Fixed emoji name lookup blowup

elimisteve:


src/components/chat/Message.js
```diff
@@ -21,8 +21,29 @@ class Message extends Component {
         // $3 == .png
         // emoji.data[$2][3][0] == smile
         // return '![:smile:](/static/img/emoji/apple/64/1f604.png)'
-        const twoSimple = ($2).split('-')[0];
-        return '![:' + emoji.data[twoSimple][3][0] + ':](' + $1 + $2 + $3 + ')';
```
elimisteve: `emoji.data[twoSimple][3][0]` was blowing up for some values of `twoSimple`.

elimisteve:
@Moe-Shoman Unfortunately `let emojiName = emojiNameArray[3][0] || 'emoji'` will thrown an exception if `emojiNameArray` is empty, hence the checks.  Or did I misunderstand you?

Moe-Shoman:
@elimisteve wouldn't it default to 'emoji' if the first value returns `undefined`?

elimisteve:
@Moe-Shoman If `emojiNameArray[3]` is `undefined` then `emojiNameArray[3][0]` will blow up, since `undefined[0]` doesn't make sense, as `undefined`  isn't a container, obviously :-)

elimisteve:
Note to self @elimisteve: add unit tests! Exhaustively go through the emojis and ensure that only `:male-sign:` and `:female-sign:` are buggy.

PR #182, opened 2017-10-05
Emoji auto-suggestions, reduxify chat.

Moe-Shoman:


elimisteve:
@Moe-Shoman Thanks!

src/index.js
```diff
@@ -3,19 +3,18 @@ import ReactDOM from 'react-dom';
...
 import './static/sass/main.scss';
 import './static/fonts/Lato.ttf';
 import './static/audio/notification_gertz.wav';
 
-const epicMiddleware = createEpicMiddleware(rootEpic)
+const epicMiddleware = createEpicMiddleware(rootEpic);
 const enhancer = compose(
   applyMiddleware(epicMiddleware),
-  ReduxDevTools.instrument()
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
```
elimisteve: @Spetastian We couldn't get `ReduxDevTools.instrument()` to work with the Redux Dev Tools Chrome extension.  Is there something special we need to do to be able to see Redux state in that extension while using `ReduxDevTools.instrument()`?

src/reducers/chatReducer.js
```diff
@@ -25,62 +25,134 @@ const initialState = {
...
+        suggestions: action.suggestions.slice(0, 25),
+        suggestionWord: state.message.slice(state.suggestionStart),
+        highlightedSuggestion: 0
+    }) : state;
+
+  case 'CHAT_ADD_SUGGESTION':
+  const beforeSuggestion = state.message.slice(0, state.suggestionStart);
+  const afterSuggestion = state.message.slice(state.suggestionStart)
+  const formattedSuggestion = afterSuggestion.replace(state.suggestionWord, action.suggestion)
+    return Object.assign({}, state, {
```
elimisteve: @Moe-Shoman Indent these 3 lines, please (129-131)

src/reducers/chatReducer.js
```diff
@@ -25,62 +25,134 @@ const initialState = {
...
+      message: ''
+    });
+
+  case 'CHAT_TOGGLE_PICKER':
+    return Object.assign({}, state, {
+      showEmojiPicker: !state.showEmojiPicker
+  });
+
+  case 'CHAT_ADD_EMOJI':
```
elimisteve: @Moe-Shoman `CHAT_PICKER_ADD_EMOJI`?

src/components/chat/AutoSuggest.js
```diff
@@ -0,0 +1,46 @@
...
+    </ul>
+  );
+
+  return (
+    <div className="suggestions-container">
+      <div className="suggestions-header">
+        Emoji matching <strong>"{chat.suggestionWord}"</strong>
+        <span className="header-help">
+	  	    <strong>↑</strong><strong>↓ </strong> to navigate
+  	  		<span className="inline-margin"><strong> ↵ </strong> to select</span>
```
elimisteve: @Moe-Shoman #tabfreelife

src/epics/chatEpics.js
```diff
@@ -53,6 +56,22 @@ import 'rxjs/add/observable/from';
...
+      return action$.ofType('CHAT_MESSAGE_UPDATE')
+      .map((msg) => showSuggestions(cursor.cursorIndex, msg.message))
+      .takeUntil(action$.ofType('CHAT_STOP_SUGGESTIONS'))
+      .catch((err) => console.error(err))
+    })
+
+const suggestionEpic = (action$) =>
+  action$.ofType('CHAT_ADD_SUGGESTION')
+    .map(() => stopSuggestions())
+    .catch((err) => console.error(err))
```
elimisteve: @Spetastian Do we have a global error helper we can use for this sort of situation?

Spetastian: No, not right now. Also this will break if this observable emits an error. It should always map to an action that redux handles. IMO, we should always tell the user if something goes wrong. 

The lest we should do right now is logging the error to the console and display an alert that something went wrong and that more information can be found in the console.

src/index.js
elimisteve: @Spetastian Ping :-) . Curious how `ReduxDevTools.instrument()` is supposed to work and whether is does something more than `window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()` for us.

src/epics/chatEpics.js
elimisteve: :+1: sounds good

src/index.js
Spetastian: Yeah, this is actually my misconception as to how setup using Redux devtools. The documentation was a bit vague. 

This here works as a basic setup and integration for using Redux devtools extension and it has a fallback if the extension is missing: https://github.com/cryptag/leapchat/pull/185

elimisteve:
@Moe-Shoman One thing I thought about earlier: when coding, be sure to disable caching --

1. Open JS console
2. Click "Network" tab
3. Check "Disable cache"

elimisteve:
The odd behavior we saw on Saturday/Sunday, where the same JS seemed to behave differently in each of our browsers, really did seem like a bug/inconsistency in Chrome/Chromium -- it was working on your machine the way the spec says it should -- but disabling cache will help ensure that Chrome is running the very latest code on your box.

PR #184, opened 2017-10-10
Fixes auto scroll down when new messages arrive

Spetastian:


elimisteve:
@Moe-Shoman This works fine for me, but it was mostly on your machine where we saw this happening.  Can you run this branch on your machine to ensure that this fixes the issue?  Here's how:

```
git remote add spetastian git@github.com:Spetastian/leapchat.git
git fetch --all
git checkout spetastian/message-auto-scroll
```

then `npm run dev` as usual.

elimisteve:
@Spetastian LGTM thus far :-)

elimisteve:
@Spetastian LGTM thus far :-)

src/epics/helpers/ChatHandler.js
```diff
@@ -54,6 +55,7 @@ class ChatHandler {
           .map(JSON.parse)
           .do(contents => {
             this.wsMessageSubject.next({
+              id: guid.create(),
```
Spetastian: @elimisteve FYI. In order to easier distinguish messages when deciding wether to scroll down or not i just create a unique ID locally until we have a better solution

src/components/chat/MessageBox.js
```diff
@@ -14,66 +14,15 @@ class MessageBox extends Component {
...
-  hasNewMessages(prevProps){
-    return prevProps.messages.length !== this.props.messages.length;
+  scrollToBottom = () => {
+    playNotification();
+    this.messagesEnd && this.messagesEnd.scrollIntoView({behavior: "smooth"});
   }
 
-  scrollToBottom(){
-    this.messagesEnd.scrollIntoView({behavior: "smooth"});
+  // Separate function to set reference because of https://reactjs.org/docs/refs-and-the-dom.html#caveats
```
elimisteve: @Spetastian Interesting!...

src/components/chat/MessageBox.js
```diff
@@ -14,66 +14,15 @@ class MessageBox extends Component {
...
-    this.setState({
-      notifiedIds: notifiedIds.concat(newMessageIds)
-    });
-  }
 
-  hasNewMessages(prevProps){
-    return prevProps.messages.length !== this.props.messages.length;
+  scrollToBottom = () => {
+    playNotification();
+    this.messagesEnd && this.messagesEnd.scrollIntoView({behavior: "smooth"});
```
Moe-Shoman: @Spetastian if you remove `{behavior: "smooth"}` it automatically starts at the bottom on page load.

elimisteve: @Spetastian Want to make this tweak, then I'll merge?

src/components/chat/MessageList.js
```diff
@@ -5,8 +5,26 @@ import Message from './Message';
 
 class MessageList extends Component {
 
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextProps.messages.length > 0 
+      && nextProps.messages
+        .map(m => m.id)
+        .concat(this.props.messages.map(m => m.id))
+        .reduce((acc, id) => {
+          if(!acc.indexOf(id) === -1){
```
elimisteve: @Spetastian Is this the same as `acc.indexOf(id) !== -1`?

src/components/chat/MessageList.js
```diff
@@ -5,8 +5,26 @@ import Message from './Message';
...
+            acc.push(id)
+          }
+          return acc;
+        }, [])
+        .length !== this.props.messages;
```
elimisteve: @Spetastian :+1: I see what you did here... If the number of old messages + the number of new messages !== the number of messages already shown, re-render.

Users can't delete messages right now, which simplifies things.  Later we'll remove out-of-sight messages from the DOM for performance reasons, at which point maybe we should compare the IDs of the newest incoming message and the latest newest rendered message to determine whether a re-render should occur?

Spetastian: Yeah, somehow we should compare the the subset of the messages that are visible to the user with the "same" subset from the new props.

But this logic is just to determine if we should scroll down. Later we may also have a property that tells the time when there were new messages and instead we compare that prop to determine if we should scroll down. Or show a button that says "new messages, scroll down" or whatever.

PR #190, opened 2017-10-12
Added util function for extracting message metadata from tags

Spetastian:


src/utils/chat.js
```diff
@@ -1,16 +1,14 @@
...
+  return {
+    isChatmessage: tags.includes('type:chatmessage'),
+    isUserStatus: tags.includes('type:userstatus'),
+    isPicture: tags.includes('type:picture'),
+    isRoomName: tags.includes('type:roomname'),
+    isRoomDescription: tags.includes('type:roomdescription'),
+    from: tagByPrefixStripped(tags, 'from:'),
+    to: tagByPrefixStripped(tags, 'to:'),
+    status: tagByPrefixStripped(tags, 'status:')
+  }
```
elimisteve: @Spetastian How about also adding `tags: tags`?  Or should we wait until we really need it?

Spetastian: IMO, we dont need the tags elsewhere. Instead we extend this parsing here. If we need another type or property from the tags we add it to the object returned here and extract it from the tags. And if this gets to bloated somewhere in the future, we refactor it then.

src/utils/chat.js
```diff
@@ -1,16 +1,14 @@
...
-      from: tagByPrefixStripped(msgRow.plaintags, 'from:'),
-      to: tagByPrefixStripped(msgRow.plaintags, 'to:'),
-      tags: msgRow.plaintags
-    };
-  });
-
-  return messages.sort(sortRowByCreated);
+export function extractMessageMetadata(tags) {
+  return {
+    isChatmessage: tags.includes('type:chatmessage'),
```
elimisteve: @Spetastian `isChatMessage`, with a capital `M`, for consistency?

Spetastian: Lol, yes.

src/utils/chat.js
elimisteve: Hmm, okay that works.  This solution is less dynamic, and requires that we do more manual work -- namely adding new specific cases whenever new tags may be added -- rather than just inspecting the tags downstream, but it should generally be a bit faster, since we do 1 check up front rather than doing string comparisons in an array.

Spetastian: Yeah, this will do the job right now. If more cases are added rapidly then we need to rethink how we do this. 

IMO, we should avoid preemptive optimization since we do not know the scope of all possible cases and this is not a big thing to refactor later on.

Spetastian:
@elimisteve I just realized that we are using `Array.includes` here as well. I fixed it in this PR.

src/utils/chat.js
elimisteve: Good point :-)

PR #256, opened 2023-02-12
docker compose rework

jimmcgaw:
What's in this PR:

1. A Dockerfile was created for building the leapchat server as a go binary docker image.
2. A Dockerfile was created for building a frontend image.
3. Updated the docker compose file to build the frontend image, go server image, as well as the two database images.
4. Updated the docker-compose file database image steps to include passing the PostgREST config and running the init script in the `postgres` container.

WIP: have the sql init shell script run in the postgres container when it boots up.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,16 +1,18 @@
 FROM golang:1.19-alpine
 
-WORKDIR /github.com/cryptag/leapchat/miniware
+RUN mkdir -p /app
 
-WORKDIR /github.com/cryptag/leapchat
+WORKDIR /app/github.com/cryptag/leapchat/miniware
+
+WORKDIR /app/github.com/cryptag/leapchat
 
 COPY miniware/* miniware/
 COPY *.go .
 COPY go.* .
 COPY LICENSE.md .
 
-RUN go build -o /leapchat
+RUN go build
 
 EXPOSE 8080
 
-CMD ["/leapchat"]
+CMD ["./leapchat"]
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,16 @@
+FROM golang:1.19-alpine
+
+WORKDIR /github.com/cryptag/leapchat/miniware
+
+WORKDIR /github.com/cryptag/leapchat
+
+COPY miniware/* miniware/
+COPY *.go .
+COPY go.* .
+COPY LICENSE.md .
+
+RUN go build -o /leapchat
+
+EXPOSE 8080
+
+CMD ["/leapchat"]
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,18 @@
+
+start:
+	docker compose up -d
+
+stop:
+	docker compose down
+
+restart:
+	docker compose restart
+
+build:
+	docker compose build
+
+buildup:
+	docker compose up -d --build
+
+cleanv:
+	docker compose down -v
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
 
+shell:
+	docker compose exec app /bin/sh
+
 start:
 	docker compose up -d
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
     postgres:
         image: postgres:latest
         ports:
-            - 5432:5433
+            - 5433:5432
         environment:
             - POSTGRES_PASSWORD=superuser
             - POSTGRES_USER=superuser
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,17 +1,34 @@
-version: '3.1'
+version: '3.4'
 
 services:
+    app:
+        build:
+            context: .
+            dockerfile: ./Dockerfile.dev
+        ports:
+            - 8080:8080
+        depends_on:
+            - postgrest
+        volumes:
+            - ./build:/build
     postgres:
         image: postgres:latest
         ports:
-            - 127.0.0.1:5432:5432
+            - 5432:5433
         environment:
             - POSTGRES_PASSWORD=superuser
             - POSTGRES_USER=superuser
             - POSTGRES_DB=leapchat
         volumes:
             - ./_docker-volumes/postgres:/var/lib/postgresql/data
+            #- ./db:/docker-entrypoint-initdb.d/
+        healthcheck:
+            test: ["CMD-SHELL", "pg_isready"]
+            interval: 5s
+            timeout: 60s
+            retries: 10
     postgrest:
+        command: postgrest /app/postgrest.conf
         image: postgrest/postgrest:latest
         ports:
             - 3000:3000
@@ -23,5 +40,8 @@ services:
             PGDATABASE: leapchat
             PGSCHEMA: public
             DB_ANON_ROLE: postgres
+        volumes:
+            - ./db/:/app
         depends_on:
-            - postgres
+            postgres:
+                condition: service_healthy
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
         depends_on:
             - postgrest
         volumes:
-            - ./build:/build
+            - ./build:/app/github.com/cryptag/leapchat/build
     postgres:
         image: postgres:latest
         ports:
--- a/leapchat.go
+++ b/leapchat.go
@@ -25,9 +25,9 @@ func init() {
 }
 
 func main() {
-	httpAddr := flag.String("http", "127.0.0.1:8080",
+	httpAddr := flag.String("http", "0.0.0.0:8080",
 		"Address to listen on HTTP")
-	httpsAddr := flag.String("https", "127.0.0.1:8443",
+	httpsAddr := flag.String("https", "0.0.0.0:8443",
 		"Address to listen on HTTPS")
 	domain := flag.String("domain", "", "Domain of this service")
 	iframeOrigin := flag.String("iframe-origin", "",
@@ -38,7 +38,7 @@ func main() {
 	flag.Parse()
 
 	if *onionPush {
-		*httpAddr = "127.0.0.1:5001"
+		*httpAddr = "0.0.0.0:5001"
 		BUILD_DIR = "public"
 	}
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 node_modules/
+build/


.gitignore
```diff
@@ -10,4 +10,7 @@ static/lib/
 _docker-volumes/
 
 # Editors
-.vscode/
\ No newline at end of file
+.vscode/
+
+# Mac OS X Bullshit
```
elimisteve: :ok_hand:

Makefile
```diff
@@ -0,0 +1,3 @@
+
+build:
+	docker compose build chatserver
```
elimisteve: `Makefile`, nice; I like this a lot and have been meaning to start using them for Go projects.

elimisteve:
Nice PR! Thank you.

I think I have a permanent solution to both the `miniware/` issue and a temporary fix for the React dependency hell issue that's good enough for the foreseeable future.

If I solve the former, shall I just push it to this branch? Or shall I branch off of it so you can take a look?

jimmcgaw:
> Nice PR! Thank you.
> 
> I think I have a permanent solution to both the `miniware/` issue and a temporary fix for the React dependency hell issue that's good enough for the foreseeable future.
> 
> If I solve the former, shall I just push it to this branch? Or shall I branch off of it so you can take a look?

Feel free to add whatever to this branch. ⚡

Dockerfile.dev
```diff
@@ -3,9 +3,10 @@ FROM golang:1.19-alpine
 WORKDIR /app
 
 RUN mkdir -p github.com/cryptag/leapchat/miniware
-COPY miniware/miniware.go github.com/cryptag/leapchat/miniware
 
-COPY miniware/* .
+WORKDIR /app/github.com/cryptag/leapchat
+
+COPY miniware/* miniware/
```
jimmcgaw, line 9: 🏇

elimisteve:
@jimmcgaw Just pushed! Got the Go build working. `docker compose up`  now almost works :smile:

docker-compose.yml
```diff
@@ -23,5 +35,7 @@ services:
             PGDATABASE: leapchat
```
elimisteve, line 40: @jimmcgaw How about we make `PGPORT` 5433 so those of us running Postgres on our host can still run LeapChat for local dev?

elimisteve:
> @jimmcgaw Just pushed! Got the Go build working. `docker compose up` now almost works smile

@jimmcgaw Want to take it from here with the dockerization? While I work on dependency hell sheeit.

docker-compose.yml
elimisteve: Want to do this?

Dockerfile.dev
elimisteve: :boom: :rocket:

frontend/package.json
```diff
@@ -32,60 +32,65 @@
...
-    "@babel/register": "^7.18.9",
-    "@webpack-cli/generators": "^3.0.1",
-    "babel-plugin-system-import-transformer": "^4.0.0",
+    "babel-eslint": "^7.2.3",
+    "babel-plugin-system-import-transformer": "^3.1.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-preset-latest": "^6.24.1",
+    "babel-preset-react": "^6.24.1",
+    "babelify": "^7.3.0",
```
jimmcgaw: Just so I'm clear: we do not want to upgrade the packages in the frontend the latest just yet?

The point of the `frontend` directory was to get the packages on the latest and get a passing build that way. It looks like you reverted some packages in the root dir and got it passing that way.

If we don't want to upgrade now, I'll revert the `frontend` directory to the way I had it and put it on a feature branch for later use.

elimisteve: I'm sorry, I thought your goal was to dockerize a working version of the app ASAP, so yeah I reverted the deps.

> Just so I'm clear: we do not want to upgrade the packages in the frontend the latest just yet?

That's what I'm thinking, yes. But if we run into issues adding features then we'll have no choice and need to update the deps and let's revisit this. Let's try to add features first!

We can immediately use the docker work you did here and that I just tweaked :slightly_smiling_face: .

> I'll revert the `frontend` directory to the way I had it and put it on a feature branch for later use.

That sounds perfect, thank you :pray: .

elimisteve:
What do you think is the best way to make it so we can modify LeapChat source files on our host but have the dev server that's running in Docker update when something changes? Shall we attach a volume (that is the folder containing the source code on the host) to the frontend container from the outside?

elimisteve:
> The database steps might need tweaking. This is just a starting point for doing local development with docker.

Excellent.

I believe the main reason why the Docker config didn't work in the past is because `db/init_sql.sh` is never executed inside the `postgres` container.

Just created an issue for this: https://github.com/cryptag/leapchat/issues/257 ...and added it to this new milestone: https://github.com/cryptag/leapchat/milestone/7

Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
...
+COPY *.go .
+COPY go.* .
+COPY LICENSE.md .
+
+RUN go build -o /leapchat
+
+CMD ["/leapchat"]
```
jimmcgaw: @elimisteve I had this in my own branch. Force-pushed to resolve the conflict with your last commit.

Since you're making changes, I'm going to stop working for now. :)

elimisteve: Hey Jim, didn't mean to colonize your branch! You finished creating `upgrade_frontend` fast and I should have let you come back to this branch unmolested.

These changes look great.

elimisteve:
I think we need to create `Dockerfile.postgrest.dev` and have it copy `db/postgrest.conf` into the `postgrest` container when building it.

elimisteve:
Created an issue for that: https://github.com/cryptag/leapchat/issues/259

elimisteve:
Alright, this PR and branch are all yours :sweat_smile: .

In my mind the last remaining things needed to do here in the milestone I made for dockerization: https://github.com/cryptag/leapchat/milestone/7

jimmcgaw:
> I think we need to create `Dockerfile.postgrest.dev` and have it copy `db/postgrest.conf` into the `postgrest` container when building it.

I don't think we need to do that. I create a volume that mounts the /db directory into the container, and a command that runs the `postgrest` command with the conf file, where it's mounted in the container, when it starts up. If that doesn't work I'll tweak it.

jimmcgaw:
> > The database steps might need tweaking. This is just a starting point for doing local development with docker.
> 
> Excellent.
> 
> I believe the main reason why the Docker config didn't work in the past is because `db/init_sql.sh` is never executed inside the `postgres` container.
> 
> Just created an issue for this: #257 ...and added it to this new milestone: https://github.com/cryptag/leapchat/milestone/7

Yep. I started a volume mount in postgres container that mounts to `/docker-entrypoint-initdb.d/`, which is the directory that a postgres container will run scripts out of, if they are present, when initialized. I haven't confirmed that this works yet.

jimmcgaw:
> What do you think is the best way to make it so we can modify LeapChat source files on our host but have the dev server that's running in Docker update when something changes? Shall we attach a volume (that is the folder containing the source code on the host) to the frontend container from the outside?

For the frontend stuff, yes. 

For the go, the ideal workflow isn't clear to me yet. I don't think I've ever worked with Docker with a compiled (ie not interpreted or JIT compiled) language. Any ideas there?

docker-compose.yml
```diff
@@ -1,17 +1,38 @@
-version: '3.1'
+version: '3.4'
 
 services:
+    frontend:
+        build:
+            context: .
+            dockerfile: ./Dockerfile.frontend.dev
+        ports:
+            - 5001:5001
```
jimmcgaw: TODO: maybe not necessary, depending on how frontend content is served locally?

docker-compose.yml
jimmcgaw: I think we can leave PGPORT as 5432, but bind it to 5433 on the host, so the container doesn't conflict.

The default 5432 should still be usable by other containers in the network.

docker-compose.yml
```diff
@@ -1,17 +1,38 @@
...
+    chatserver:
+        build:
+            context: .
+            dockerfile: ./Dockerfile.dev
+        ports:
+            - 8080:8080
+        depends_on:
+            - postgrest
```
jimmcgaw, line 11: @elimisteve I sorta skimmed your thread from last night in the email, re: networking. I don't think we need to use networking mode = "host" to expose things. The default bridge network should be sufficient.

I exposed port 8080 in the golang container and that allowed me to load via `127.0.0.1:8080`.

package.json
```diff
@@ -45,9 +45,9 @@
     "js-sha512": "^0.3.0",
     "markdown-it": "^8.4.2",
     "prop-types": "^15.7.2",
-    "react": "^18.2.0",
+    "react": "^15.7.0",
```
jimmcgaw: Thank you for sorting out the dependency hell. :)

docker-compose.yml
jimmcgaw: Ah I see. There's a flag (`onionpush`) that serves the content from `public/` directory on 5001, but that's in the golang container, not here.

docker-compose.yml
elimisteve: You're right that it shouldn't be needed. I couldn't get port 8080 to appear open from my host without it, but I'm glad it worked for you and hopefully it works for me after I reboot so that my user is thoroughly in the `docker` group, and/or [toggle my VPN on or off](https://stackoverflow.com/a/57648759/197160), or something like that :sweat_smile: lol. #software

docker-compose.yml
elimisteve: Sounds good

docker-compose.yml
```diff
@@ -1,17 +1,38 @@
...
     postgres:
         image: postgres:latest
         ports:
-            - 127.0.0.1:5432:5432
+            - 5432:5433
```
elimisteve: @jimmcgaw I think these should be reversed, yes? To avoid port collisions on the server.

docker-compose.yml
elimisteve: To be clear: let's do what you said :+1: ; I'll figure out the networking shit on my end, and if I can't then I'll just run it on the host, which I've already got set up.

docker-compose.yml
jimmcgaw: Yep, got these backwards. 🙅

elimisteve: :+1:

Dockerfile.dev
```diff
@@ -0,0 +1,18 @@
...
+COPY miniware/* miniware/
+COPY *.go .
+COPY go.* .
+COPY LICENSE.md .
+
+RUN go build -o /leapchat
+
+EXPOSE 8080
+
+CMD ["/leapchat"]
```
jimmcgaw: @elimisteve So you're right, `npm run dev` just does a build to the `/build` directory. The leapchat binary just serves content directly from there.

I removed the Docker frontend container / file. I don't think it makes sense right here, given the current workflow.

Instead, in the immediate-term, I think we can create a volume in this container that binds `build` to a directory next to the binary in the container? `npm start` would then run the `watch-build` loop and update the `build` dir in the go container.

elimisteve: @jimmcgaw Removing that other container and mounting `build` into the Go container sounds perfect.

The only detail I'd add is: instead of putting the `leapchat` Go binary at `/leapchat`, let's put it next to the other *.go files just so that `leapchat` is running in the same environment everywhere: in Docker, on my laptop, and on the production server. I mention that here because this implies putting the `build` directory right next to `leapchat` and thus inside the `/app/github.com/cryptag/leapchat` directory.

Sound good?

EDIT: I expanded on and improved on this idea below.

elimisteve: In that case I think we just need this:

```dockerfile
FROM golang:1.19-alpine

RUN mkdir -p /app/github.com/cryptag/leapchat/miniware

WORKDIR /app/github.com/cryptag/leapchat/

COPY miniware/* miniware/
COPY *.go .
COPY go.* .
COPY LICENSE.md .

RUN go build

EXPOSE 8080

CMD ["./leapchat"]
```

And then to mount `./build` as `/app/github.com/cryptag/leapchat/build` inside the container. (I believe `/app/github.com/cryptag/leapchat/build` doesn't need to and indeed probably shouldn't exist inside the container, since mounting a volume to that directory would overwrite it or maybe not work?)

Makefile
```diff
@@ -0,0 +1,18 @@
...
+	docker compose restart
+
+build:
+	docker compose build
+
+buildup:
+	docker compose up -d --build
+
+cleanv:
+	docker compose down -v
```
elimisteve: (Copied from the email I just sent since this is better context for it --)

How about if I [some time soon, but not in this branch right now] add `make release` to cut a new (tagged, versioned) release, and `make deploy` to create a production tarball then scp it to the server?

elimisteve (commented):
Looking pretty great! See comments.

docker-compose.yml
elimisteve: That's true.

Marking this as resolved since the frontend container is no longer needed.

Dockerfile.dev
elimisteve: Actually, if this works we can make it more like you suggested above, with a simpler directory structure in the container:

```dockerfile
FROM golang:1.19-alpine

RUN mkdir -p /leapchat/miniware

WORKDIR /leapchat

COPY miniware/* miniware/
COPY *.go .
COPY go.* .
COPY LICENSE.md .

RUN go build

EXPOSE 8080

CMD ["./leapchat"]
```

elimisteve (commented):
Oops, I meant for these to be one-off comments, not part of a new review

leapchat.go
```diff
@@ -25,9 +25,9 @@ func init() {
 }
 
 func main() {
-	httpAddr := flag.String("http", "127.0.0.1:8080",
+	httpAddr := flag.String("http", "0.0.0.0:8080",
```
jimmcgaw, line 28: This change allowed me to get a simple "404 not found" blank error page when accessing localhost:8080. So, I think it just doesn't know how to find the static `build/` files.

Makefile
```diff
@@ -1,4 +1,7 @@
 
+shell:
+	docker compose exec app /bin/sh
+
```
elimisteve, line 4: Dope

leapchat.go
elimisteve: Yup yup, makes sense

Dockerfile.dev
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.19-alpine
+
+RUN mkdir -p /app
+
+WORKDIR /app/github.com/cryptag/leapchat/miniware
+
+WORKDIR /app/github.com/cryptag/leapchat
```
elimisteve, line 7: @jimmcgaw I believe we want the 3 lines above into these 2, so that our working directory exists and so we don't have the additional `WORKDIR`:

```dockerfile
RUN mkdir -p /app/github.com/cryptag/leapchat/miniware
WORKDIR /app/github.com/cryptag/leapchat
```

elimisteve: Then I think this PR is ready to merge! Is it working well for you locally?

elimisteve:
> For the go, the ideal workflow isn't clear to me yet. I don't think I've ever worked with Docker with a compiled (ie not interpreted or JIT compiled) language. Any ideas there?

The Go binary will need to change so infrequently, and it compiles so quickly, that I say we don't worry about having anything watch the file system for changes to .go files; let's just manually rebuild the Go container when we need to.

Sound good?

Dockerfile.dev
elimisteve: Oh yeah there needs to be a bit more done: https://github.com/cryptag/leapchat/milestone/7

bhagwatis:
New to docker and postgrest. I have installed new version of docker.  Containers installed posgresql, pgadmin4 and postgrest.  Postgrest does not start in the container nor am i able to ping that specific port.  Can someone please help me here?

PR #265, opened 2023-02-12
Unit test for minilock parse method

jimmcgaw:
Still learning the go.

I can't assert that this test is useful, being a project n00b here. :)

`go test -run "^TestParseMinilockID$"`

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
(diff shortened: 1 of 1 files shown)


elimisteve:
@jimmcgaw Looks great! FYI commands like this will also run the test:

    go test -run TestParseMinilockID

or even

    go test -run TestParseMinilo

(it matches by prefix, I believe).

PR #273, opened 2023-02-14
React cleanup / renaming; add button by username to edit username

jimmcgaw:


src/components/App.js
```diff
@@ -108,45 +100,49 @@ class App extends Component {
...
-        <main className="encloser">
-
-          {pincodeRequired && <PincodeModal
-            showModal={pincodeRequired}
-            onSetPincode={this.onSetPincode}
-            onCloseModal={this.onClosePincodeModal} />}
-
-          {displaySettings && <UsernameModal
-            previousUsername={previousUsername}
+          <Header
```
elimisteve: This is tiny, but I think this is indented one level too far?

src/reducers/chatReducer.js
```diff
@@ -26,7 +26,7 @@ const initialState = {
   previousUsername: previousUsername,
   status: '',
   messages: [],
-  statuses: [],
+  statuses: {},
```
elimisteve, line 29: Is this correct?

I think I made it an array instead of an object to avoid the race condition where the client deletes a user's status right before that (still-connected) client sends a new status.

(...Not that those statuses are actually being deleted correctly, as you may have noticed; if someone ever connected during your current visit, the UI shows them as still being connected, even if that's not true. Maybe the big refactor I mentioned earlier broke those deletions...)

src/components/App.js
jimmcgaw: Yep, I'll run the linter.

src/reducers/chatReducer.js
jimmcgaw: Not sure if this is correct, but this is consistent with the object the app ends up creating for statuses in the app code.

I also expected it to be an array, but the actual variable being passed around was an object with username keys.

src/components/App.js
elimisteve: Looks good, thanks

src/reducers/chatReducer.js
elimisteve: Gotcha; makes sense.

elimisteve: Checking out this branch (in both senses of the term) in a sec.

elimisteve:
Oops, I just detected a bug; commenting here since it's in this code.

@jimmcgaw For some reason when multiple users are in a room, the ` (me)` label disappears for oneself.

src/components/chat/UserList.js
```diff
@@ -59,8 +57,10 @@ class UserList extends Component {
           {viewing.map((username, i) => {
             return (
               <li key={i}>
-                <ViewingUserIcon />
-                {username}
+                <ViewingUserIcon
+                  username={username}
+                  isCurrentUser={username === currentUsername}
```
elimisteve, line 62: @jimmcgaw I know what's happening -- ^^this^^ sort of line needs to be added to the 2 other statuses! `viewing` is the green dot, so when their dot is yellow, `isCurrentUser` doesn't get set and so ` (me)` doesn't show up.

elimisteve: Fixed in https://github.com/cryptag/leapchat/pull/277

PR #278, opened 2023-02-15
Privacy tweaks: rel="nofollow noreferrer noopener"

elimisteve:
And disable autocomplete on username field

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -107,41 +107,40 @@ class App extends Component {
 
     return (
       <div id="page">
-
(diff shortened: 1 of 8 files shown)


elimisteve:
The only thing unique to this PR now is `autoComplete="off"`; closing this PR because of that and because our force-pushes messed something up with the relationship between these two branches.

elimisteve:
We can add `autoComplete="off"` elsewhere as it's not really related to an existing PR anyway

PR #279, opened 2023-02-15
Message search

jimmcgaw:
message-seach

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6002,6 +6002,11 @@
       "integrity": "sha512-OW2r4sQ0sI+z5ckEt5c1Tri4xTgZwYDxpE54eqWlQloQRoWtXjqt9udJ5Z4dSv7wK+nfFI7FRXyCpBSft+gpFw==",
       "dev": true
     },
+    "minisearch": {
+      "version": "6.0.1",
+      "resolved": "https://registry.npmjs.org/minisearch/-/minisearch-6.0.1.tgz",
+      "integrity": "sha512-Ly1w0nHKnlhAAh6/BF/+9NgzXfoJxaJ8nhopFhQ3NcvFJrFIL+iCg9gw9e9UMBD+XIsp/RyznJ/o5UIe5Kw+kg=="
+    },
     "minizlib": {
       "version": "2.1.2",
       "resolved": "https://registry.npmjs.org/minizlib/-/minizlib-2.1.2.tgz",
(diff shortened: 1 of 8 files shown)


elimisteve:
:fire: :fire: :fire:

src/components/modals/MessageSearchModal.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  render() {
+    const {
+      isVisible,
+      searchResults,
+      isLoadingResults,
+      username } = this.state;
+
+    const sortedResults = searchResults.sort((a, b) => a.score > b.score);
```
elimisteve: @jimmcgaw Hey hey.

This line mutates `searchResults` so it may not make sense to create `sortedResults` as a separate variable.

jimmcgaw: Oh yeahhh, I forgot that this method sorts an array in-place. I’ll remove the extra var.

elimisteve:
1. When an emoji is used in a message that was found via search, that emoji is bigger than it should be. Please constrain its size like the existing CSS does for messages.

2. How about keeping the purple messages purple and the white ones white? I guess certain props aren't being passed to the messages in the search modal or something, so the ChatMessage objects in the modal all think they were sent by not-you?

3. In the future it'd be cool to be able to put the app into Search Mode by clicking the search icon and filtering down the chat messages "in-line" -- that is, without needing the modal at all -- but that seems like a lot of work for not a huge benefit right now.

jimmcgaw:
> * When an emoji is used in a message that was found via search, that emoji is bigger than it should be. Please constrain its size like the existing CSS does for messages.
> * How about keeping the purple messages purple and the white ones white? I guess certain props aren't being passed to the messages in the search modal or something, so the ChatMessage objects in the modal all think they were sent by not-you?
> * In the future it'd be cool to be able to put the app into Search Mode by clicking the search icon and filtering down the chat messages "in-line" -- that is, without needing the modal at all -- but that seems like a lot of work for not a huge benefit right now.

1 and 2 are super simple tweaks, of course. 3 makes me think you might have some more thought-out uses cases for search that haven’t been considered here. Let’s not merge until we chat next.

elimisteve:
> 3 makes me think you might have some more thought-out uses cases for search that haven’t been considered here.

@jimmcgaw I hadn't thought much about search until now, but I just realized: with your modal approach it'll be easier to have universal search across not just messages, but tasks and todo lists, and eventually files... I think this is what Slack does, and it works well. So let's stick with the approach you took here :ok_hand: ; very nice.

PR #281, opened 2023-02-15
Add playwright and browser tests; remove selenium stuff; update README.md

jimmcgaw:
The pointer to playwright testing library was a good one. This thing is pretty awesome.

This removes all of the selenium junk (which ran successfully) and adds playwright for browser tests. I added enough tests to more than have parity with what selenium was doing. Also updated the README where JS tests are outlined.


.github/workflows/pull_request_javascript_check.yml
```diff
@@ -16,4 +16,6 @@ jobs:
         node-version: 14.x
     - run: npm install
     - run: npm run lint
-    - run: npm test
\ No newline at end of file
+    - run: npm run mocha
+    # - run: npx playwright install-deps
+    # - run: npm test
```
jimmcgaw, line 21: Unfortunately, `localhost:8080` is not available on the stock Github ubuntu build machine. :)

elimisteve:
Looks amazing :ok_hand: . And I just remembered that this exists, making it _really_ fast to "write" new tests! https://playwright.dev/docs/codegen

jimmcgaw:
Indeed, codegen looks promising. Next steps are to figure out the best way to organize the playwright tests and how they should interact with the DOM.

Issue #199, opened 2020-04-19
Infelicity with the browser UI: the symbols by the usernames are unexplained


LinuxOnTheDesktop:
It would be good were the browser UI to explain - at least via its help button - what the symbols by the usernames mean.

For one thing: what do the coloured circles mean? I had a green circle next to one user and another green circle next to another user, though just now one of those circles changed colour to yellow. Say what? Perhaps it is to do with two of the users being on the same IP.

Also, what is the import of the little hyphens - or are they chevrons - next to the circles? They look somewhat clickable but they turn out not to be.

Further: the user may fear that closing the 'connected to server' banner will abort his/her connection. Also, at least on my computers (using Firefox - on Windows and Linux) the banner obscures a username.

PR #284, opened 2023-02-16
Add tooltips to convey status icon meaning

jimmcgaw:
This is meant to address #199.  I don't think I like this solution.

status-tooltips

### DIFF
--- a/src/components/chat/UserStatusIcons.js
+++ b/src/components/chat/UserStatusIcons.js
@@ -1,9 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+
 import FaCircle from 'react-icons/lib/fa/circle';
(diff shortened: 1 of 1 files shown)


src/components/chat/UserStatusIcons.js
```diff
@@ -13,9 +19,27 @@ export const UserStatusIcon = ({
...
+        delayShow={300}
+        delayHide={150}>
+        <FaCircle style={styleViewing} />
+      </OverlayTrigger>
+    );
   } else if (status === 'online') {
-    statusIcon = <FaCircle style={styleOnline} />;
+    statusIcon = (
+      <OverlayTrigger
+        placement="right"
```
elimisteve, line 35: Is `top` any better?

jimmcgaw:
I slept on this one. Going to close this.

It may not be immediately clear what these icons mean, but I think most users can glean it from context and the tooltips are just clutter. I can revisit if users voice confusion.

PR #287, opened 2023-02-16
Add toolbar control for toggling audio, persist preference between sessions

jimmcgaw:
This PR introduces an option to mute the audio in the LeapChat browser window.

We haven't discussed audio design; perhaps this should be configurable in settings at a more granular level, ie "play sound for all new message", "play sound only when I am mentioned in a message", etc.) But **_at a minimum_**, the user should be able to turn it off.

What this PR does:
1. When the user first loads the window and is presented with the "Set Username" dialog, don't play the audio in the background. This triggers a JS console error.
2. Once the Username modal is dismissed, check if the user has previously set an audio-enabled preference. If so, default to that, else just enable it by default.
3. An icon on the chat toolbar that allows for toggling the audio-enabled preference. This is persisted to the App state so it affects the session immediately and is persisted to localStorage to be used in future visits / sessions.
4. Also, there are some React annotations, like prop types for components, blah blah.

toggle-audio

### DIFF
--- a/src/components/chat/MessageBox.js
+++ b/src/components/chat/MessageBox.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 
 import MessageList from './MessageList';
-import Throbber from '../general/Throbber';
 import { connect } from 'react-redux';
 import { closePicker } from '../../actions/chatActions';
 import { playNotification } from '../../utils/audio';
@@ -16,8 +15,14 @@ class MessageBox extends Component {
     };
   }
 
+  onNewMessages = () => {
+    if (this.props.isAudioEnabled){
+      playNotification();
+    }
+    this.scrollToBottom();
+  }
+
   scrollToBottom = () => {
-    playNotification();
     this.messagesEnd && this.messagesEnd.scrollIntoView();
   }
 
@@ -32,7 +37,10 @@ class MessageBox extends Component {
     return (
       <div className="message-box" onClick={closePicker}>
         <div className="message-list">
-          <MessageList onNewMessages={this.scrollToBottom} messages={messages} username={username} />
+          <MessageList
+            onNewMessages={() => this.onNewMessages()}
+            messages={messages}
+            username={username} />
         </div>
         <div style={{float: "left", clear: "both"}}
           ref={this.setMessagesEndRef}>
--- a/src/components/chat/MessageForm.js
+++ b/src/components/chat/MessageForm.js
@@ -1,10 +1,11 @@
 import React, { Component } from 'react';
-import ReactDOM from 'react-dom';
+import { PropTypes } from 'prop-types';
+import { connect } from 'react-redux';
+
 import { Button } from 'react-bootstrap';
 import FaArrowCircleRight from 'react-icons/lib/fa/arrow-circle-right';
 import FaSmileO from 'react-icons/lib/fa/smile-o';
 import { Picker } from 'emoji-mart';
-import { connect } from 'react-redux';
 import emoji from '../../constants/emoji';
 import { emojiSuggestions, mentionSuggestions } from '../../utils/suggestions';
 import {
@@ -19,6 +20,7 @@ import {
   upSuggestion,
   addSuggestion
 } from '../../actions/chatActions';
+import ToggleAudioIcon from './toolbar/ToggleAudioIcon';
 
 class MessageForm extends Component {
   constructor(props) {
@@ -132,7 +134,11 @@ class MessageForm extends Component {
 
   render() {
     const { message, showEmojiPicker } = this.props.chat;
-    const { messageUpdate, togglePicker } = this.props;
+    const {
+      messageUpdate,
+      togglePicker,
+      isAudioEnabled,
+      onSetIsAudioEnabled } = this.props;
 
     return (
       <div className="message-form">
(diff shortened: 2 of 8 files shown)


src/components/chat/MessageBox.js
```diff
@@ -16,8 +15,14 @@ class MessageBox extends Component {
     };
   }
 
+  onNewMessages = () => {
+    if (this.props.isAudioEnabled){
+      playNotification();
+    }
+    this.scrollToBottom();
```
jimmcgaw, line 22: Ancillary note: if I'm reading this correctly, I could see this becoming a nuisance if a user is scrolling back up to read past messages and they keep getting pulled back down.

src/components/chat/MessageForm.js
```diff
@@ -157,9 +163,14 @@ class MessageForm extends Component {
                 onClick={togglePicker}
               />
 
+              <ToggleAudioIcon
+                isAudioEnabled={isAudioEnabled}
+                onSetIsAudioEnabled={onSetIsAudioEnabled} />
+
               <div className="right-chat-icons"></div>
-            </div>
+            </div>
```
elimisteve: Trailing whitespace

src/components/modals/Username.js
```diff
@@ -46,6 +46,12 @@ class UsernameModal extends PureComponent {
 
     if (this.isUsernameValid(username)) {
       this.props.onSetUsername(username);
+      // set the audio to the user's previously selected preference; enable by default
+      let isAudioEnabled = JSON.parse(localStorage.getItem('isAudioEnabled'));
```
elimisteve: In case this code gets executed before `onSetIsAudioEnabled()`, how about replacing

```js
JSON.parse(localStorage.getItem('isAudioEnabled'))
```

with

```js
JSON.parse(localStorage.getItem('isAudioEnabled') || 'true')
```

so that an exception can't be thrown here?

But overall this looks great, thank you.

elimisteve (changes requested):
Muy bien, but see 2 small comments.

src/components/modals/Username.js
jimmcgaw: For my own edification, what exception are we worried about?  `localStorage.getItem` returns `null` if no key is found. AFAIK, `JSON.parse` will not have a problem parsing that to `null`.

The `|| 'true'` within `JSON.parse` is cleaner than the following few lines. 👍

src/components/chat/MessageBox.js
elimisteve: Yup, I've experienced that. We could add a "Scroll to bottom" floating overlay button in the bottom right when the user scrolls up, and not automatically scroll down when a new message comes in...

src/components/modals/Username.js
elimisteve: Maybe I'm being too paranoid but `JSON.parse('')` would throw an exception.

I wasn't sure what parsing the literal `null` would do but you're right, it parses fine even though it's not a string. Good to know!

jimmcgaw: Ah, k. `JSON.parse('')` isn't inconceivable. Good catch.

elimisteve:
@jimmcgaw Rough UX thoughts; feel free to act on whichever parts you want, or hold off till we get user feedback, or have me to them, or suggest different solutions, or suggest things stay as they are:

1. Having the button where it is makes it easy to get to, but arguably it's more intuitive for it to be in the Settings modal.

2. Let's at least have equal horizontal spacing of the icons in this bar just above the textarea, including equal spacing between the left-most icon and the left edge of the bar, methinks.

3. Having the speaker jump right when muted feels a little off? Maybe force it to stay to the left so the spacing between icons remains constant? (If this is the only icon that changes its width when clicked then in the future maybe it should remain the right-most icon in the bar. Or maybe it's more weird for it to move as new features are added. Could put a vertical separator bar to its left to help solve that? Then the mute/enable audio button will always be to the right of the separator.)

jimmcgaw:
> @jimmcgaw Rough UX thoughts; feel free to act on whichever parts you want, or hold off till we get user feedback, or have me to them, or suggest different solutions, or suggest things stay as they are:
> 
> 1. Having the button where it is makes it easy to get to, but arguably it's more intuitive for it to be in the Settings modal.

This won't be the last time I rearrange the UI, including figuring out what goes in the Settings modal. My thoughts: if we need to show max progress at the end of these couple of weeks, let's leave it visible for now.

> 2. Let's at least have equal horizontal spacing of the icons in this bar just above the textarea, including equal spacing between the left-most icon and the left edge of the bar, methinks.

I'll add some padding to the left of the emoji picker. Sound good?

> 3. Having the speaker jump right when muted feels a little off? Maybe force it to stay to the left so the spacing between icons remains constant? (If this is the only icon that changes its width when clicked then in the future maybe it should remain the right-most icon in the bar. Or maybe it's more weird for it to move as new features are added. Could put a vertical separator bar to its left to help solve that? Then the mute/enable audio button will always be to the right of the separator.)

It doesn't actually change widths, the "off" icon is just narrower than the "on" one, so it seems to jump. If there were another icon to the right of the speaker, it wouldn't shift. For now, I think the visual is helpful because it clearly lets the user know something has changed.

In the long term: the most recent font-awesome icons version has many speakers, and the mute / unmute icon possibilities are much richer, and they wouldn't shift if changed. I'd like to get us to using those instead, so would rather not mess with this too much.

In generally, happy to discuss design or UX stuff more generally so I have a sense of how to make these decisions, instead of ad-hoc. Or if you have a style guide... :)

elimisteve:
Sounds good! Pun intended... And nice work 🤠.

PR #294, opened 2023-02-18
Upgrade react-icons pkg to get newer icon versions

jimmcgaw:
Upgrades the `react-icons` package to latest, gets us better icons. Update with equivalent new ones in app.

See new "audio toggle": 😄 

toggle-audio-new-icons


jimmcgaw:
If you approve, let's merge these in before the search branch and I'll update that branch to use the new icons as well.

elimisteve:
@jimmcgaw Done!

PR #295, opened 2023-02-19
Sharing modal

jimmcgaw:
What's in this PR:

1. Create a single helper method for showing / hiding modals by their name so we reduce the number of functions we drill down as props.
2. Create a separate sharing, invite users modal that contains a means of sharing via the Web Share API, when supported, and allows for copying a link to share, as previously allowed.
3. Add a trigger to open the invite users modal to the Message toolbar and as a button under the users list.

Inviting users should be as easy and discoverable as possible. I felt it needed to be pulled out of the Settings modal where it was buried.

Demo, in desktop Chrome (does not support Web Share API) and desktop Safari (which does support Web Share API):

invite-users

### DIFF
--- a/src/components/modals/InfoModal.js
+++ b/src/components/modals/InfoModal.js
@@ -2,59 +2,60 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Modal } from 'react-bootstrap';
 
-export default class InfoModal extends Component {
-  constructor(props){
-    super(props);
-  }
+const InfoModal = ({ isVisible, onToggleModalVisibility }) => {
 
-  render() {
-    return (
-      <Modal show={this.props.showModal} onHide={this.props.onToggleInfoModal}>
-        <Modal.Header closeButton>
-          <h3>Welcome to LeapChat!</h3>
-          <h4>LeapChat: encrypted, ephemeral, in-browser chat.</h4>
-        </Modal.Header>
-        <Modal.Body>
-          <p>Just visit <a href="https://www.leapchat.org/" target="_blank">leapchat.org</a> and a new, secure chat room will instantly be created for you. And once you're in, just link people to that page to invite them to join you!</p>
-          <h3>
-            Why LeapChat?
-          </h3>
-          <p>
-            You shouldn't have to sacrifice your privacy and personal information just to chat online. Slack, HipChat, and others make you create an account with your email address, their software doesn't encrypt your messages (they can see everything), and the messages last forever unless you manually delete them.
-            In contrast, LeapChat <em>does</em> encrypt your messages (even we can't see them!), <em>doesn't</em> require you to hand over your email address, and messages last for a maximum of 90 days (this will soon be configurable to a shorter duration).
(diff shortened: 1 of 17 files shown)


src/components/modals/InfoModal.js
```diff
@@ -2,59 +2,60 @@ import React, { Component } from 'react';
...
+        <h3>
+          Can I type markdown in my messages?
+        </h3>
+        <p>
+          <em>Yup!</em> To learn about Markdown syntax, like surrounding words with **double asterisks** to make them <b>bold</b>, or with _underscores_ to make them <i>italicized</i>, <a href="https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet" target="_blank">check out this guide</a>.
+        </p>
+      </Modal.Body>
+    </Modal>
+  );
+};
```
jimmcgaw, line 54: This is just adjusting indentation when converting this to a functional component.

src/components/modals/SharingModal.js
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  return (
+    <div>
+      <Modal show={isVisible} onHide={onClose}>
+        <Modal.Header closeButton>
+          <Modal.Title>Invite to Chat</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          { navigator.share && <div>
```
jimmcgaw, line 42: On show this where supported.

src/components/modals/SharingModal.js
```diff
@@ -0,0 +1,88 @@
...
+
+const onCopyShareLink = (e) => {
+  navigator.clipboard.writeText(window.location.href);
+};
+
+const onShareLink = (e) => {
+  navigator.share({
+    url: window.location.href,
+    title: "LeapChat",
+    text: "Join me on LeapChat"
```
jimmcgaw, line 16: Executive callout: if we enable sharing this way, this text might be the first thing new users potentially see when invited to the platform. Placeholder text; verbiage feedback humbly requested.

src/components/modals/SharingModal.js
```diff
@@ -0,0 +1,88 @@
...
+                  <FaShareAlt size={15} />
+                </Button>
+              </OverlayTrigger>
+            </form>
+          </div>
+
+        </Modal.Body>
+        <Modal.Footer>
+          
+        </Modal.Footer>
```
jimmcgaw, line 76: TODO: nuke or do something with this.

src/components/modals/SharingModal.js
```diff
@@ -0,0 +1,88 @@
...
+                delayHide={150}>
+                <Button className="icon-button" bsStyle="primary">
+                  Copy to Clipboard
+                  <FaShareAlt size={15} />
+                </Button>
+              </OverlayTrigger>
+            </form>
+          </div>
+
+        </Modal.Body>
```
jimmcgaw, line 73: Also a place for sharing an "Admin" link, once that's ready. :D

src/components/modals/SharingModal.js
elimisteve: Sounds good

PR #298, opened 2023-02-20
Offer use of in-browser speech to text transcript if available.

jimmcgaw:
I wanted to build this for myself, more than anything else.

Success:

speech-to-text-success

Error:

speech-to-text-error

### DIFF
--- a/src/components/chat/MessageForm.js
+++ b/src/components/chat/MessageForm.js
@@ -21,6 +21,7 @@ import {
   addSuggestion
 } from '../../actions/chatActions';
 import ToggleAudioIcon from './toolbar/ToggleAudioIcon';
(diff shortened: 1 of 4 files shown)


src/components/chat/MessageForm.js
```diff
@@ -132,14 +133,23 @@ class MessageForm extends Component {
     this.messageInput = input;
   }
 
+  onMessageUpdate = (e) => {
+    const message = e.target.value;
+    this.props.messageUpdate(message);
+  }
```
jimmcgaw, line 139: Even if we don't ship this feature, I'd like to update this at some point. IMHO, the redux actions should not be coupled with browser event handlers directly.

PR #301, opened 2023-02-21
Upgrade frontend fixups

jimmcgaw:


src/components/layout/Info.js
```diff
@@ -15,9 +15,9 @@ const Info = ({ onToggleModalVisibility }) => {
   
   return (
     <div className="info">
-      {/* <OverlayTrigger placement="bottom" overlay={infoTooltip} delayHide={150} delayShow={300}>
-        <FaInfoCircle onClick={onClose} size={25}/>
-      </OverlayTrigger> */}
+      {/* <OverlayTrigger placement="bottom" overlay={infoTooltip} delayHide={150} delayShow={300}> */}
+      {/* </OverlayTrigger> */}
```
elimisteve, line 19: @jimmcgaw Leaving this in as a reminder to replace this?

jimmcgaw: Aye. Going to see about incrementally upgrading react-bootstrap and see if we can get the same component without compatibility issues with react. If that won't work, then I'll find some other tool in this version to accomplish the same thing.

elimisteve: Cool, sounds good. I don't know if this is any good, but... it does exist :sweat_smile: https://github.com/ReactTooltip/react-tooltip/

elimisteve: Additional options: https://reactscript.com/best-tooltip-component/

But using `react-bootstrap` sounds good to me as long as it works well with the latest version of React.

PR #304, opened 2023-02-23
Extract initial connection auth dance from the epics into module

jimmcgaw:
There are benefits to observables, arguably. I don't see any benefit to having this initial auth handshake with the server being contained in the epics and being done via observables.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import {
   setUsername,
+  initAuth,
   initConnection,
   initChat
 } from '../store/actions/chatActions';
@@ -42,7 +43,7 @@ class App extends Component {
   }
 
   connectIfNeeded() {
-    if (!this.props.pincodeRequired && this.props.shouldConnect) {
+    if (!this.props.pincodeRequired && this.props.shouldConnect){
       this.onInitConnection();
     }
   }
@@ -57,6 +58,8 @@ class App extends Component {
 
 
   onInitConnection(pincode='') {
+    this.props.initAuth();
+
     const urlHash = document.location.hash + pincode;
     initiateSessionAndConnect(
       this.props.initConnection,
@@ -139,8 +142,18 @@ const mapStateToProps = (reduxState) => {
 const mapDispatchToProps = (dispatch) => {
   return {
     initChat: () => dispatch(initChat()),
-    initConnection: (pincode) => dispatch(initConnection(pincode)),
-    initApplication: () => dispatch(initApplication()),
+    initAuth: () => dispatch(initAuth()),
+    initConnection: ({
+      authToken,
+      secretKey,
+      mID,
+      isNewRoom
+    }) => dispatch(initConnection({
+      authToken,
+      secretKey,
+      mID,
+      isNewRoom
+    })),
     setUsername: (username) => dispatch(setUsername(username)),
   };
 };
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -6,6 +6,8 @@ import {
   initChat
 } from '../store/actions/chatActions';
 
+import { initiateSessionAndConnect } from '../utils/sessions';
+
 import Header from './layout/Header';
 
 import ChatContainer from './chat/ChatContainer';
@@ -41,8 +43,30 @@ class App extends Component {
 
   connectIfNeeded() {
     if (!this.props.pincodeRequired && this.props.shouldConnect) {
-      this.props.initConnection();
+      this.onInitConnection();
     }
+  }
+
+  onSetPincode = (pincode = "") => {
+    if (!pincode || pincode.endsWith("--")) {
+      this.onError('Invalid pincode!');
+      return;
+    }
+    this.onInitConnection(pincode);
+  };
+
+
+  onInitConnection(pincode='') {
+    const urlHash = document.location.hash + pincode;
+    initiateSessionAndConnect(
+      this.props.initConnection,
+      this.createWebSession,
+      urlHash,
+    );
+  }
+
+  createWebSession(passphrase) {
+    document.location.hash = "#" + passphrase;
   }
 
   onToggleModalVisibility = (modalName, isVisible) => {
@@ -59,12 +83,16 @@ class App extends Component {
     });
   };
 
-  onSetPincode = (pincode = "") => {
-    if (!pincode || pincode.endsWith("--")) {
-      this.onError('Invalid pincode!');
-      return;
-    }
-    this.props.initConnection(pincode);
+  onSetIsAudioEnabled = (isAudioEnabled) => {
+    this.setState({
+      isAudioEnabled: isAudioEnabled
+    });
+    localStorage.setItem("isAudioEnabled", isAudioEnabled);
+  };
+
+  onSetUsername = (username) => {
+    this.props.setUsername(username);
+    this.onToggleModalVisibility('username', false);
   };
 
   render() {
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -83,18 +83,6 @@ class App extends Component {
     });
   };
 
-  onSetIsAudioEnabled = (isAudioEnabled) => {
-    this.setState({
-      isAudioEnabled: isAudioEnabled
-    });
-    localStorage.setItem("isAudioEnabled", isAudioEnabled);
-  };
-
-  onSetUsername = (username) => {
-    this.props.setUsername(username);
-    this.onToggleModalVisibility('username', false);
-  };
-
   render() {
     const {
       username,
--- a/src/components/chat/ChatContainer.js
+++ b/src/components/chat/ChatContainer.js
@@ -37,8 +37,6 @@ ChatContainer.propTypes = {
 
 const mapStateToProps = (reduxState) => {
   return {
-    messages: reduxState.chat.messages,
-    username: reduxState.chat.username,
     suggestions: reduxState.chat.suggestions,
   };
 };
--- a/src/utils/sessions.js
+++ b/src/utils/sessions.js
@@ -1,0 +1,60 @@
+import { getEmail, getPassphrase } from '../utils/encrypter';
+import miniLock from '../utils/miniLock';
+
+// TODO: will be different host from a mobile device, probably if (!window)
+const authUrl = `${window.location.origin}/api/login`;
+
+
+async function connectWithAuthRequest(initiateConnection, mID, secretKey, isNewPassphrase) {
+  const response = await fetch(authUrl, {
+    method: "GET",
+    headers: {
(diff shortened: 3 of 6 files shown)


src/components/App.js
```diff
@@ -67,10 +69,34 @@ class App extends Component {
...
+      this.onError('Invalid pincode!');
+      return;
+    }
+    // this.props.initConnection(pincode);
+    this.onInitConnection(pincode);
+  };
+
+
+  onInitConnection(pincode='') {
+    const urlHash = document.location.hash + pincode;
```
jimmcgaw, line 63: Step 1. The first issue was that the very start of the observable auth flow had a dependency on the browser. This just injects the browser value into flow. When it's not the browser, ie a mobile device, it will come from a navigation screen parameter or something.

src/utils/sessions.js
```diff
@@ -0,0 +1,60 @@
...
+    isNewPassphrase
+  } = getPassphrase(urlHash);
+
+  // 2. Get email based on passphrase
+  let email = getEmail(passphrase);
+  // 3. Decrypt to get key pair
+  miniLock.crypto.getKeyPair(passphrase, email, (keyPair) => {
+    miniLock.session.keys = keyPair;
+    miniLock.session.keyPairReady = true;
+    let mID = miniLock.crypto.getMiniLockID(keyPair.publicKey);
```
jimmcgaw, line 51: Step 2: initial entry point, do the minilock stuff.

src/utils/sessions.js
```diff
@@ -0,0 +1,60 @@
...
+    // 4. When we have keypair, create session on device
+    if (isNewPassphrase) {
+      createWebSession(passphrase);
```
jimmcgaw: Step 3: callback hook to do something on the device. In this case, just a browser, but might be something else on mobile, etc.

src/utils/sessions.js
```diff
@@ -0,0 +1,60 @@
...
+  miniLock.crypto.decryptFile(message, mID, secretKey,
+    function (fileBlob, saveName, senderID) {
+      const reader = new FileReader();
+      reader.addEventListener("loadend", () => {
+        const authToken = reader.result;
+        initiateConnection({
+          authToken,
+          secretKey,
+          mID,
+          isNewRoom: isNewPassphrase,
```
jimmcgaw, line 26: Step 4. Do the auth dance, get the token, and send that to the init connection action in redux.

src/actions/chatActions.js
```diff
@@ -13,8 +13,13 @@ export const CHAT_USERNAME_SET = 'CHAT_USERNAME_SET';
...
 
-export const initConnection = (pincode = '') =>
-  ({ type: CHAT_INIT_CONNECTION, pincode });
+export const initConnection = ({
+  authToken, secretKey, mID, isNewRoom
+}) =>
+  ({
+    type: CHAT_INIT_CONNECTION,
+    authToken, secretKey, mID, isNewRoom
+  });
```
jimmcgaw: Step 5. This is the action dispatched by Step 4, when doing the auth. Now we enter the observable flow, when it receives this action type.

src/epics/chatEpics.js
```diff
@@ -73,98 +67,34 @@ const suggestionEpic = (action$) =>
...
-      'X-Minilock-Id': mID
-    },
-    method: 'GET'
-  };
-  return settings;
-}
+};
 
 const initConnectionEpic = (action$) =>
   action$.ofType(CHAT_INIT_CONNECTION)
```
jimmcgaw: Step 6: This listens for this action to be triggered. We enter this flow with the action parameters. We pass those, first thing, to the initChatHandlerConnection method in here.

src/epics/chatEpics.js
```diff
@@ -73,98 +67,34 @@ const suggestionEpic = (action$) =>
...
-        });
-
-        reader.readAsText(fileBlob);
-      });
+const initChatHandlerConnection = ({ authToken, secretKey, mID, isNewRoom }) => {
+  return chatHandler.initConnection({
+    authToken,
+    secretKey: secretKey,
+    mID
   });
```
jimmcgaw: Step 7. Initialize the chat handler's connection. Since `chatHandler` is a shared global variable in this module, this is as far as we can go, without ripping all of the other observable stuff out of here too. This returns an observable.

src/epics/chatEpics.js
```diff
@@ -73,98 +67,34 @@ const suggestionEpic = (action$) =>
...
-          console.error(error);
-          return Observable.from([
-            alertWarning('Lost connection to chat server. Trying to reconnect...'),
-            disconnected()
-          ]);
-        })
-    )
+    .mergeMap(initChatHandlerConnection)
+    .mergeMap((isNewRoom) =>
+      Observable.merge(
```
jimmcgaw: Step 8: with `chatHandler` initialized and connected, do all the same stuff we did before.

jimmcgaw:
@elimisteve There will be a conflict between this and other open PR that reworks major chunks of the frontend. In terms of ease of resolving conflicts, I'd rather merge the other one first.

src/components/App.js
```diff
@@ -67,10 +69,34 @@ class App extends Component {
 
   connectIfNeeded() {
     if (!this.props.pincodeRequired && this.props.shouldConnect) {
-      this.props.initConnection();
+      // this.props.initConnection();
```
elimisteve: Can this line be removed?

src/epics/chatEpics.js
elimisteve: So much more readable! Bravo 👏

src/utils/sessions.js
```diff
@@ -0,0 +1,60 @@
...
+const authUrl = `${window.location.origin}/api/login`;
+
+
+async function connectWithAuthRequest(initiateConnection, mID, secretKey, isNewPassphrase) {
+  const response = await fetch(authUrl, {
+    method: "GET",
+    headers: {
+      'X-Minilock-Id': mID
+    }
+  });
```
elimisteve, line 14: `fetch` ftw 👌

src/components/App.js
jimmcgaw: Ah, yep. Last push was some minor cleanup of three lines, including removing this.

src/utils/sessions.js
elimisteve: @jimmcgaw In case more things are added to the body of `createWebSession()`, should we perhaps add `await` here (and make this callback we're creating here and passing to `miniLock.crypto.getKeyPair(...)` `async`)?

jimmcgaw: I can add that.

src/utils/sessions.js
```diff
@@ -0,0 +1,60 @@
...
+  // 2. Get email based on passphrase
+  let email = getEmail(passphrase);
+  // 3. Decrypt to get key pair
+  miniLock.crypto.getKeyPair(passphrase, email, async (keyPair) => {
+    miniLock.session.keys = keyPair;
+    miniLock.session.keyPairReady = true;
+    let mID = miniLock.crypto.getMiniLockID(keyPair.publicKey);
+    // 4. When we have keypair, create session on device
+    if (isNewPassphrase) {
+      await createWebSession(passphrase);
```
jimmcgaw, line 54: Step 3. Etc

src/components/chat/ChatContainer.js
```diff
@@ -37,8 +37,6 @@ ChatContainer.propTypes = {
 
 const mapStateToProps = (reduxState) => {
   return {
-    messages: reduxState.chat.messages,
-    username: reduxState.chat.username,
```
jimmcgaw, line 41: Small clean up from the state PR. Piggybacking onto this.

src/components/App.js
```diff
@@ -59,12 +83,16 @@ class App extends Component {
...
+  onSetIsAudioEnabled = (isAudioEnabled) => {
+    this.setState({
+      isAudioEnabled: isAudioEnabled
+    });
+    localStorage.setItem("isAudioEnabled", isAudioEnabled);
+  };
+
+  onSetUsername = (username) => {
+    this.props.setUsername(username);
+    this.onToggleModalVisibility('username', false);
```
jimmcgaw: Snuck in on bad merge with develop. Will remove.

src/utils/sessions.js
elimisteve: Gracias :grin: :man_dancing:

src/components/App.js
elimisteve: Feel free to merge this in once you do! Lookin' good :sunglasses:

jimmcgaw:
@elimisteve I just added a commit to this, and you'll want to take a look. I was seeing a subtle bug. Sometimes, when I connected, my messages were double posting into the message box. On page reload, the duplicate messages disappeared.

Why? The app, via redux state, is one big finite state machine. Dispatching actions handles the state transitions. (You're probably aware of this. Giving context.)

After initializing the chat, the app goes through these state transitions:

`(shouldConnect) == initConnection ==> (connecting) == connectionInitiated ==> (connected)`

The app continuously checks its own state (at least, on every re-render) to see if it's in shouldConnect state. If so, it initiates an initConnection transition. Previously, the init connection updated the state to "connecting" immediately, so any re-renders wouldn't trigger a duplicate connection. But, now that we moved the initConnection out of the observable flow and into its own module, there was a significant delay between init'ing the connection process and the state being updated. The state now gets updated at the very END of the flow, once the auth handshake has completed.

This caused a subtle race condition. If the user updated the state of the app during the auth handshake (say, by setting their username) it would look and find the app in the `shouldConnet` state, even though the connection had been initiated. So another auth / connection was initiated. The result? You'd end up with double connections handling requests on the `chatHandler`. This would result in double messages being appended to the local `messages` list, but not double requests to the server. (So a page reload would remove the duplicate messages.)

I updated the state machine and added an additional state, and transition. It now looks like this:

`(shouldConnect) == initAuth ==> (authenticating) == initConnection ==> (connecting) == connectionInitiated ==> (connected)`

`initAuth` action dispatch immediately transitions the app out of `shouldConnect` state, so even if there are re-renders, it won't initiate a second connection.

src/components/App.js
```diff
@@ -123,8 +142,18 @@ const mapStateToProps = (reduxState) => {
 const mapDispatchToProps = (dispatch) => {
   return {
     initChat: () => dispatch(initChat()),
-    initConnection: (pincode) => dispatch(initConnection(pincode)),
-    initApplication: () => dispatch(initApplication()),
```
jimmcgaw, line 127: `initApplication` looks like it's no longer used, or even defined, anywhere.

elimisteve:
Excellent; thanks for adding the `connecting` state

elimisteve:
Or, rather, the `CHAT_INIT_AUTH` state.

Issue #200, opened 2020-04-19
Three small problems with the 'connected to server' banner - and a single solution for all three


LinuxOnTheDesktop:
1) The user may fear that closing the 'connected to server' banner will abort his/her connection.

2) At least on my computers (using Firefox - on Windows and Linux) the banner obscures a username.

3) After one closes the banner, it pops up again after a while. Is there any point to that? The user who wants the banner gone must close it again.

**Suggestion to solve all three problems**: _turn the banner into text, i.e. remove its close icon and possibly its colouring, and put it in the left pane_. In that pane, it will not obscure anything (thus solving 2) and so will not need to be closable (solving problems 2 and 3).

LinuxOnTheDesktop:
Also:

4) it is unclear whether the browser is merely cosmetic. That is: _will closing it sever the connection to the server?_ Actually -and if I may say so - _that is a question that I need answered before I start using the site in earnest._

elimisteve:
Closing the notification does not sever anything.

LinuxOnTheDesktop:
elimsteve: right; thanks.

jimmcgaw:
@elimisteve I'm going to second some of the usability issues noted in here. This alert shows up every time you switch back to the browser / tab after a short time of being idle or away, it obscures the top navigation on mobile web (for me, this happens in mobile Safari), you have to dismiss it every single time, the dismiss button is small and easy to miss clicking. So, I vote we alter this somehow.

One idea: as suggested above, we could make this a text notification that renders for a short duration, then auto-dismisses, and which doesn't obscure any vital content.

Another idea: we could have a "connection status" icon that is clearly visible somewhere, and which switches between clearly differentiated "connected" and "disconnected" status icons as the connection changes.

I vote for the latter, because we should be saving the "notifications" / "alerts" real estate for other more important use cases.

Thoughts?

elimisteve:
@jimmcgaw Interesting ideas, and I agree the notification is annoying. How about we just don't trigger this notification? Then if we feel like we're missing context/visual cues that we're connected, we do as you just described, where we indicate connection status (if that doesn't make the UI feel too busy), or show a notification toast.

(If only for future reference, I will note that react-toastify is really great.)

How's that sound?

jimmcgaw:
I'm good to nuke it for the time being. I would think that a "connection error" message should show if the user tries to do anything and the connection is bad, instead of telling them they're connected. 

[...]

Issue #139, opened 2017-07-27
On new room: don't make user click "New room created!"; should timeout


elimisteve:
Feature idea from Marty!

elimisteve:
@MartyTheeMartian For LeapChat, want to just move to React Toastify?

PR #307, opened 2023-02-25
Welcome loading and connecting dialogue

jimmcgaw:
welcome-message

A simple welcome message and "connecting" indicator dialogue when the user first loads the page. Also renders an alert that indicates a new room was created or an existing one joined.

This is meant in part to address #200 

Also meant to address a slight usability issue, in that the username dialog form isn't responsive to user input until the connection has been established anyway. This would almost certainly be frustrating for quick-fingered users experiencing initial connection latency.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -16,6 +16,9 @@ import ChatContainer from './chat/ChatContainer';
 import PincodeModal from './modals/PincodeModal';
 import UsernameModal from './modals/Username';
 
+// evaluate on initial render only, not on every re-render.
+const isNewRoom = Boolean(!document.location.hash);
+
 class App extends Component {
   constructor(props) {
     super(props);
@@ -90,7 +93,11 @@ class App extends Component {
     const {
       username,
       pincodeRequired,
-      previousUsername } = this.props;
+      previousUsername,
+      authenticating,
+      connecting,
+      connected,
+    } = this.props;
 
     let showUsernameModal = this.state.modals.username.isVisible;
     showUsernameModal = !pincodeRequired && (showUsernameModal || username === '');
@@ -112,7 +119,11 @@ class App extends Component {
           previousUsername={previousUsername}
           username={username}
           isVisible={showUsernameModal}
+          isNewRoom={isNewRoom}
           setUsername={this.props.setUsername}
+          authenticating={authenticating}
+          connecting={connecting}
+          connected={connected}
           onToggleModalVisibility={this.onToggleModalVisibility} />}
 
         <main className="encloser">
@@ -136,6 +147,9 @@ const mapStateToProps = (reduxState) => {
     previousUsername: reduxState.chat.previousUsername,
     pincodeRequired: reduxState.chat.pincodeRequired,
     shouldConnect: reduxState.chat.shouldConnect,
+    connecting: reduxState.chat.connecting,
+    connected: reduxState.chat.connected,
+    authenticating: reduxState.chat.authenticating,
   };
 };
 
--- a/src/components/modals/Username.js
+++ b/src/components/modals/Username.js
@@ -2,7 +2,7 @@ import React, { useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
-import { Modal, Button } from 'react-bootstrap';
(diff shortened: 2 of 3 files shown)


src/components/modals/WelcomeLoadingModal.js
```diff
@@ -0,0 +1,59 @@
...
+        <h3>Welcome to LeapChat!</h3>
+      </Modal.Header>
+      <Modal.Body>
+        {isNewRoom && <Alert
+          bsStyle="success">
+          New Room Created!{' '}<FaFrog />
+        </Alert>}
+        {!isNewRoom && <Alert
+          bsStyle="success">
+          Successfully joined room!{' '}<FaFrog />
```
elimisteve: The frog is cute but people will think it's a [hat-tip to the far right](https://www.theguardian.com/world/2017/may/08/pepe-the-frog-creator-kills-off-internet-meme-co-opted-by-white-supremacists) :disappointed: . Pick something else at least as cool?

jimmcgaw: Oh yeah, I forgot about Pepe. Yah, I'll nuke it.

elimisteve:
@jimmcgaw Thanks for doing this, but honestly I think this makes the app load time feel significantly slower because of the 2 modals instead of 1, plus the light that flashes between modals kind of hurts my eyes; could be annoying for people. Instead of a modal to show progress, how about we use `react-toastify` to tell them they're connected to the server, then that message fades after maybe 2 seconds? Toasts could even show the different stages of connecting that you laid out (but as toasts rather than a modal!).

LeapChat issue: https://github.com/cryptag/leapchat/issues/139

The `react-toastify` settings we used for a previous code base; some variant of this seems appealing to me: https://github.com/PursuanceProject/pursuance/blob/179b502c1a4fe65f4468beb7cc9797b66e151abf/src/components/Content/TaskHierarchy/TaskHierarchy.js#L144 plus https://github.com/PursuanceProject/pursuance/blob/179b502c1a4fe65f4468beb7cc9797b66e151abf/src/components/Content/TaskHierarchy/TaskHierarchy.js#L34

jimmcgaw:
`react-toastify` looks cools, and I'm game to add it. I'd want to consider using that when we add "message failed to send" notifications, on transmission errors.

But, this was not intended to solve the problem of "inform the user about their connection status". When the username modal first comes up, it remains unresponsive to submitting until the connection has succeeded. If there's latency in establishing the connection, the set username button might seem broken. There is an alert in place, but it's in the background, partially obscured by the modal, so that doesn't help clarify the situation. I think `react-toastify` would have the same issue in this case.

This also makes it clear, right from the first, whether a room was created or if you're joining one. It does this before we ask for a username. But, admittedly, if this is important, then it might go away (connection is established) before they get a chance to read it.

If two modals isn't great, we could alternatively: put the "loading" indicator onto the username form, and show the username form, but have the "Set Username" button disabled with a loading indicator, so it's clear that they cannot submit it until the thing that is happening is done. Any messaging from the "Welcome" modal (or any other stuff) could be put onto the username form, so they can read it at their leisure.

I don't particularly care if we solve this problem; but, I think it is a problem that merits consideration. I'll set to draft and we can discuss at some point, if you ever wish to.

elimisteve:
> When the username modal first comes up, it remains unresponsive to submitting until the connection has succeeded. If there's latency in establishing the connection, the set username button might seem broken. There is an alert in place, but it's in the background, partially obscured by the modal, so that doesn't help clarify the situation. I think `react-toastify` would have the same issue in this case.

Good point.

> If two modals isn't great, we could alternatively: put the "loading" indicator onto the username form, and show the username form, but have the "Set Username" button disabled with a loading indicator, so it's clear that they cannot submit it until the thing that is happening is done. Any messaging from the "Welcome" modal (or any other stuff) could be put onto the username form, so they can read it at their leisure.

Great idea! That seems clean and cohesive :ok_hand: .

elimisteve:
Very nice :+1: . Merging!

The initial modal is starting to look a bit busy, and there are old issues with the "Generate Random Username" button not really being aligned with anything, but I think we can solve both issues by doing what Effective does, where to the right of the username input field there's a :game_die: the user can click to get a new random username.

PR #308, opened 2023-02-25
CapacitorJS testing WIP

jimmcgaw:
This feels very close to working, for development mode. Right now the issue is that the app, when running in an emulator, will not reach out to the localhost:8080. I've tried several workarounds from the docs, I haven't found the solution yet.


elimisteve:
> This feels very close to working, for development mode. Right now the issue is that the app, when running in an emulator, will not reach out to the localhost:8080. I've tried several workarounds from the docs, I haven't found the solution yet.

Meaning the app tries to talk to `localhost:8080` inside the emulator rather than `localhost:8080` on the host?

elimisteve:
@jimmcgaw Can conditional logic be set so that, when running in an emulator, [10.0.2.2:8080](https://stackoverflow.com/a/9777798) is reached out to instead of localhost:8080 ?

server.go
```diff
@@ -74,6 +74,7 @@ func ProductionServer(srv *http.Server, httpsAddr, domain string, manager *autoc
 
 func Login(m *miniware.Mapper, pgClient *PGClient) func(w http.ResponseWriter, req *http.Request) {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		//w.Header().Set("Access-Control-Allow-Origin", "*")
```
jimmcgaw, line 77: This is _not_ the solution to the local emulator access issue. I don't think.

src/store/epics/chatEpics.js
```diff
@@ -137,7 +137,6 @@ const ownUserStatusEpic = (action$, store) =>
         .map(() => userStatusSent())
     )
     .catch(error => {
-      console.error(error);
```
jimmcgaw, line 207: I thought I was cleaning up my own message here.

capacitor.config.json
```diff
@@ -0,0 +1,19 @@
...
+    "CapacitorHttp": {
+      "enabled": true
+    }
+  },
+  "server": {
+    "cleartext": "true",
+    "allowNavigation": [
+      "192.168.1.247:8080/*",
+      "localhost:8080/*",
+      "leapchat.org/*"
```
jimmcgaw, line 16: 🙏 📿

elimisteve: Sweet! Should probably add `www.leapchat.org/*`, I'd think?

server.go
elimisteve: kk

capacitor.config.json
elimisteve: And 10.0.2.2 -- https://developer.android.com/studio/run/emulator-networking

elimisteve: `"10.0.2.2:8080/*"`, I suppose

PR #309, opened 2023-02-26
Alerts optionally dismiss after specified duration

jimmcgaw:
Optional parameter to the show alert pipeline that makes alerts (optionally) auto-dismissing after a specified duration.

Intended to at least partially address #200 

Oh and apparently #139

### DIFF
--- a/src/store/epics/helpers/ChatHandler.js
+++ b/src/store/epics/helpers/ChatHandler.js
@@ -181,7 +181,7 @@ class ChatHandler {
     reader.readAsArrayBuffer(fileBlob);
   };
 
-  initConnection = ({ mID, secretKey, authToken }) => {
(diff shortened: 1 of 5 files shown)


src/store/epics/helpers/ChatHandler.js
```diff
@@ -195,7 +195,7 @@ class ChatHandler {
       this.ws.onmessage = this.onWsMessage;
       this.ws.onopen = (event) => {
         event.target.send(this.authToken);
-        observer.next();
+        observer.next(isNewRoom);
```
jimmcgaw, line 198: If we emit this here, then `isNewRoom` is exposed to the place in the epic where we show "New room created". (Unintentionally borked previously.)

PR #318, opened 2023-03-01
decouple chat action from the browser events api

jimmcgaw:
Small change extracted from https://github.com/cryptag/leapchat/pull/298

Redux actions shouldn't know about the browser event API.


elimisteve:
Looks good :ok_hand:

PR #323, opened 2023-03-01
Restore original init connection flow epic; inject browser dependencies into chat epics

jimmcgaw:
This cleans up the reworking of the init connect epic to be not be an epic. The objective there was to eliminate the direct dependency on the device from the epics themselves. This is much more easily done, with minimal potential unintended loss of functionality, via the means in this PR.

### DIFF
--- a/src/store/epics/helpers/urls.js
+++ b/src/store/epics/helpers/urls.js
@@ -1,0 +1,14 @@
+
+let hostname;
+if (typeof window.location !== "undefined") {
+  hostname = window.location.origin;
+} else {
+  // hard-coded for mobile device scenarios
+  hostname = "http://locahost:8080";
+}
+
+export const authUrl = `${hostname}/api/login`;
+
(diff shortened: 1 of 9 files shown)


src/store/epics/helpers/urls.js
```diff
@@ -0,0 +1,14 @@
+
+let hostname;
+if (typeof window.location !== "undefined") {
```
elimisteve, line 3: `window` is defined on mobile, but not `window.location`?

elimisteve: I'd think that `window` wouldn't be defined outside the browser.

jimmcgaw: `window` is defined in a react native environment. It's possible we end up with a different mobile scenario (say, not CapacitorJS) and will have to tweak this.

elimisteve: Good to know, thanks

PR #324, opened 2023-03-01
redux store dependencies out of dev dep section

jimmcgaw:
If you `npm install` LeapChat source as a JS dependency in a react-native app, the following libraries have been listed as dependencies, instead of dev deps, in order for the redux store to properly run itself on the device.


elimisteve:
Makes sense. If this PR continues to just move dependencies between the dev v. non-dev categories, feel free to merge it whenever you like; no need for me to review :+1: .

PR #326, opened 2023-03-02
Capacitor: Android: LeapChat login and WS connection work!!! :tada:

elimisteve:
leapchat-android-capacitor-ITWORKS-Screenshot_2023-03-02_01-20-46

The emoji image is broken but WHO CARES!!! All these work now:

- Talk to Go backend from Capacitor at all
- Get auth token from `/api/login` (and turn into `Blob` to pass to miniLock) => Successful login!
- Establish WS connection with Go backend


src/utils/sessions.js
```diff
@@ -6,31 +6,40 @@ import { CapacitorHttp } from '@capacitor/core';
 // TODO: will be different host from a mobile device, probably if (!window)
 let authUrl = `${window.location.origin}/api/login`;
 
-if (Capacitor.isNativePlatform()){
-  authUrl = "http://192.168.1.247:8080/api/login";
+if (Capacitor.isNativePlatform()) {
+  authUrl = "http://10.0.2.2:8080/api/login";
```
elimisteve, line 11: We should probably create global constants like `NATIVE_BASE_URL` or whatever, but I've hardcoded this for now.

src/store/epics/chatEpics.js
```diff
@@ -28,8 +28,13 @@ import {
 import ChatHandler from './helpers/ChatHandler';
 import { combineEpics } from 'redux-observable';
 import createDetectVisibilityObservable from './helpers/createDetectPageVisibilityObservable';
+import { Capacitor } from '@capacitor/core';
+
+let wsUrl = `${window.location.origin.replace('http', 'ws')}/api/ws/messages/all`;
+if (Capacitor.isNativePlatform()) {
```
elimisteve, line 34: I did what was quickest to get something working, but we should probably do this conditional once, in one place, and define what are effectively global constants which we then export, then import in places like this.

But... IT WORKS!!!

elimisteve:
Closing this because @jimmcgaw already merged these commits into `capacitorjs`

PR #331, opened 2023-03-06
QR code on invite screen

jimmcgaw:
qr-code

### DIFF
--- a/src/components/modals/SharingModal.js
+++ b/src/components/modals/SharingModal.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Capacitor } from '@capacitor/core';
 import { Clipboard } from '@capacitor/clipboard';
@@ -6,6 +6,10 @@ import { Clipboard } from '@capacitor/clipboard';
 import { Modal, Button, OverlayTrigger, Tooltip } from 'react-bootstrap';
 import { FaShareAlt } from 'react-icons/fa';
 import { FaShareAltSquare } from 'react-icons/fa';
+
+import { FaQrcode } from 'react-icons/fa';
+
+import QRCode from "react-qr-code";
 
 
 const getWindowLocationHref = () => {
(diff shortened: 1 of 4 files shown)


src/components/modals/SharingModal.js
```diff
@@ -7,6 +7,10 @@ import { Modal, Button, OverlayTrigger, Tooltip } from 'react-bootstrap';
 import { FaShareAlt } from 'react-icons/fa';
 import { FaShareAltSquare } from 'react-icons/fa';
 
+import { FaQrcode } from 'react-icons/fa';
+
+import QRCode from "react-qr-code";
```
jimmcgaw, line 12: The animated lib didn't have its ES6 / import interface documented.

elimisteve: Sounds good :+1:

src/components/modals/SharingModal.js
```diff
@@ -42,6 +46,11 @@ const SharingModal = ({
   isVisible,
   onClose
 }) => {
+
+  const [showQRCode, setShowQRCode] = useState(false);
+
+  const inviteLink = getWindowLocationHref();
```
elimisteve: @jimmcgaw To avoid race conditions please run `getWindowLocationHref()` as late as possible. On mobile, the app loads with `window.location.href === 'http://www.leapchat.org'`, with no URL hash :frowning_face: .

src/components/modals/SharingModal.js
```diff
@@ -75,6 +84,16 @@ const SharingModal = ({
...
+            
+          </div>
+          <hr />
+          <h3>QR Code</h3>
+          <Button className="icon-button" variant="primary" onClick={() => setShowQRCode(true)}>
+            Generate QR Code
+            <FaQrcode size={15} />
+          </Button>
+          <div className="qr-code">
+            {showQRCode && <QRCode value={inviteLink} />}
```
elimisteve, line 100: ...which is to say: please tweak this to say `{showQRCode && <QRCode value={getWindowLocationHref()} />}`

src/components/modals/SharingModal.js
elimisteve: https://github.com/cryptag/leapchat/pull/331/files#r1125968319

src/components/modals/SharingModal.js
jimmcgaw: Did it slightly differently, but done.

